PR #266, opened 2025-06-30
Fix edge-to-edge on three-button nav bar

Goooler:
| Before | After |
|--------|--------|
| before | after |


tunjid (approved):
Thank you!

PR #269, opened 2025-07-04
Evaluate Metro for DI

tunjid:


data/src/commonMain/kotlin/com/tunjid/heron/data/di/DataModule.kt
```diff
@@ -148,50 +161,48 @@ abstract class DataComponent(
...
-        @Provides get() = this
 
+    @SingleIn(DataScope::class)
+    @Binds
     val SnapshotWriteQueue.bind: WriteQueue
-        @DataScope
-        @Provides get() = this
 
+    @SingleIn(DataScope::class)
+    @Binds
```
tunjid: Seems having multiple binds methods with the same name causes a compile error.

PR #365, opened 2025-08-25
Use type-safe version catalogs accessors in precompiled script plugins

Goooler:


build-logic/convention/src/main/kotlin/ext/KspExt.kt
```diff
@@ -1,58 +0,0 @@
...
-        project.dependencies {
-            add(
-                configurationName = "ksp${
-                    targetName.replaceFirstChar {
-                        if (it.isLowerCase()) it.titlecase(
-                            Locale.getDefault()
-                        ) else it.toString()
-                    }
-                }",
-                dependencyNotation = project.versionCatalog.findLibrary("kotlin-inject-compiler")
```
Goooler, line 44: `kotlin-inject` has been replaced by Metro. Seems no need to preserve `configureKsp` for now.

build-logic/convention/build.gradle.kts
```diff
@@ -33,4 +33,7 @@ dependencies {
     implementation(libs.kotlin.serializationPlugin)
     implementation(libs.google.devtools.kspPlugin)
     implementation(libs.metro.gradlePlugin)
+
+    // TODO: https://github.com/gradle/gradle/issues/15383
+    implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
```
tunjid, line 38: Given this is a workaround, and may break in newer gradle versions, is it safer to just jeep using the non type-safe strings for now?

build-logic/convention/src/main/kotlin/ext/KspExt.kt
tunjid: Sounds good to me, I forgot to remove it completely after migration

build-logic/convention/build.gradle.kts
Goooler: I believe this workaround is stable, but either way is good to me. 

I addressed this PR here as I see the dependencies like `lifecycle-runtime` are not marked as referenced in IDEA, as they are string refs.

tunjid: Yeah makes sense, makes it easier to identify unused dependencies.

PR #391, opened 2025-09-03
Tidy up migration directories

tunjid:


tunjid:
Dropping the values in these tables without migrating their data was deliberate.

PR #417, opened 2025-09-05
Update SavedStateDataSource to save signed in user ids last seen tab

joelmuraguri:


data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/Preferences.kt
```diff
@@ -18,12 +18,14 @@ package com.tunjid.heron.data.core.models
...
 import com.tunjid.heron.data.core.types.FeedGeneratorUri
 import com.tunjid.heron.data.core.types.ProfileId
+import com.tunjid.heron.data.core.types.Uri
 import kotlinx.serialization.Serializable
 
 typealias ContentLabelPreferences = List<ContentLabelPreference>
 
 @Serializable
 data class Preferences(
+    val lastViewedHomeTimelineUri: Uri? = null,
```
tunjid: The `SavedState` is stored using [protocol buffers](https://protobuf.dev/) (or protobufs), so the order of fields are very important. We will need to add this as the last argument, or users will crash when they upgrade the app.

Good job adding null as the default value, this is also needed for backwards compatability.

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
```diff
@@ -70,6 +72,7 @@ class ActualHomeViewModel(
     @Suppress("UNUSED_PARAMETER")
     @Assisted
     route: Route,
+    savedStateDataSource: SavedStateDataSource,
```
tunjid: Lets keep the assisted injected fields as the last 2 arguments. While not a requirement for metro (the current dependency injection library used), it is a requirement for others.

Let the constructor be:

```kotlin
class ActualHomeViewModel(
    authRepository: AuthRepository,
    timelineRepository: TimelineRepository,
    savedStateDataSource: SavedStateDataSource,
    writeQueue: WriteQueue,
    navActions: (NavigationMutation) -> Unit,
    @Assisted
    scope: CoroutineScope,
    @Suppress("UNUSED_PARAMETER")
    @Assisted
    route: Route,
)
```

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
```diff
@@ -185,9 +195,12 @@ private fun Flow<Action.SendPostInteraction>.postInteractionMutations(
         writeQueue.enqueue(Writable.Interaction(action.interaction))
     }
 
-private fun Flow<Action.SetCurrentTab>.setCurrentTabMutations(): Flow<Mutation<State>> =
-    mapToMutation { action ->
-        copy(currentTabUri = action.currentTabUri)
+private fun Flow<Action.SetCurrentTab>.setCurrentTabMutations(
+    savedStateDataSource: SavedStateDataSource,
+): Flow<Mutation<State>> =
+    mapToManyMutations { action ->
```
tunjid: This is good, but we can do better. On flows, you can:
 * `map`: This will be called for every emission, and any suspending methods called here will run to completion, until the next emission is processed.
 * `mapLatest`: This will be called for every emission, but if a new emission comes in when a suspending call is still being processed, it will cancel the call and start working on the latest one immediately.

Since a user can swipe between tabs very, very quickly, we need to make sure we are certain the user has settled on a tab before we write it to disk. Otherwise, we risk writing to the disk very frequently and consuming unnecessary resources.

What we want to do instead, is to emit and write to memory very quickly, and write to storage later and after a delay that can be cancelled. To do this, update this method to:

```kotlin
private fun Flow<Action.SetCurrentTab>.setCurrentTabMutations(
    savedStateDataSource: SavedStateDataSource,
): Flow<Mutation<State>> = mapLatestToManyMutations { action -> 
    // Write to memory in state immediately
    emit { copy(currentTabUri = action.currentTabUri) }
    
    // Wait until we're sure the user has settled on this tab
    delay(1400.milliseconds)
    // Write to disk
    savedStateDataSource.setLastViewedHomeTimelineUri(action.currentTabUri)
}
```

By mapping to the latest, if we are still waiting to write a value to disk when a new value comes in, we promptly cancel the stale value, write the fresh value to memory, and start waiting to write this value to disk again.

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
```diff
@@ -121,10 +125,16 @@ private fun loadProfileMutations(
 private fun timelineMutations(
     scope: CoroutineScope,
     timelineRepository: TimelineRepository,
+    savedStateDataSource: SavedStateDataSource,
 ): Flow<Mutation<State>> =
     timelineRepository.homeTimelines().mapToMutation { homeTimelines ->
+
+        val lastViewed = savedStateDataSource.savedState.value
```
tunjid: Let's improve on what the AI suggested by using the `combine` operator:

```kotlin
combine(
    // We only care about the very first value emitted
    savedStateDataSource.savedState.take(1),
    timelineRepository.homeTimelines(),
    ::Pair,
) { (savedState, homeTimelines) -> 
    val tabUri = currentTabUri 
    ?: savedState.signedInUserPreferences()
        ?.lastViewedHomeTimelineUri?.takeIf { uri -> 
            homeTimelines.indexOfFirst { it.isPinned && it.uri == uri } >= 0 
        }
    ?: homeTimelines.firstOrNull()?.uri
    
    copy(
        currentTabUri = tabUri,
        ...
    )
}
```

This lets us:
* Preserve the last in memory value.
* Use the last saved value if it still exists in the pinned home timelines (the user may have update timelines in the browser)

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeScreen.kt
```diff
@@ -337,13 +332,11 @@ private fun HomeTimeline(
                     },
                 state = gridState,
                 columns = StaggeredGridCells.Adaptive(presentation.cardSize),
-                verticalItemSpacing = presentation.lazyGridVerticalItemSpacing,
+                verticalItemSpacing = 8.dp,
```
tunjid: Did you mean to change this?

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeScreen.kt
```diff
@@ -337,13 +332,11 @@ private fun HomeTimeline(
...
                 contentPadding = UiTokens.bottomNavAndInsetPaddingValues(
                     top = UiTokens.statusBarHeight + UiTokens.toolbarHeight + UiTokens.tabsHeight,
                 ),
-                horizontalArrangement = Arrangement.spacedBy(
-                    presentation.lazyGridHorizontalItemSpacing,
-                ),
+                horizontalArrangement = Arrangement.spacedBy(8.dp),
```
tunjid: Did you also mean to change this?

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
```diff
@@ -121,14 +129,27 @@ private fun loadProfileMutations(
 private fun timelineMutations(
     scope: CoroutineScope,
     timelineRepository: TimelineRepository,
+    savedStateDataSource: SavedStateDataSource,
 ): Flow<Mutation<State>> =
-    timelineRepository.homeTimelines().mapToMutation { homeTimelines ->
+    combine(
```
tunjid, line 134: This would be a little neater with a method reference:

```kotlin
    combine(
        savedStateDataSource.savedState.take(1),
        timelineRepository.homeTimelines(),
        ::Pair,
    ) 
    ```

File: feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt [hunk 8]
```diff
@@ -140,8 +161,11 @@ private fun timelineMutations(
                         timelineRepository = timelineRepository,
                     )
 
-                if (timeline.isPinned) HomeScreenStateHolders.Pinned(timelineStateHolder)
-                else HomeScreenStateHolders.Saved(timelineStateHolder)
+                if (timeline.isPinned) {
```
tunjid: Unless linting failed, I prefer simple if statements to not have braces; it reduces visual noise. I use braces when the code becomes hard to read without them.

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
```diff
@@ -121,10 +125,16 @@ private fun loadProfileMutations(
...
 ): Flow<Mutation<State>> =
     timelineRepository.homeTimelines().mapToMutation { homeTimelines ->
+
+        val lastViewed = savedStateDataSource.savedState.value
+            .signedInUserPreferences()
+            ?.lastViewedHomeTimelineUri
+
         copy(
-            currentTabUri = currentTabUri ?: homeTimelines.firstOrNull()?.uri,
+            currentTabUri = lastViewed ?: currentTabUri ?: homeTimelines.firstOrNull()?.uri,
```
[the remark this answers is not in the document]
tunjid: Resolved.

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
tunjid: Resolved.

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
tunjid: Resolved.

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
```diff
@@ -121,14 +129,27 @@ private fun loadProfileMutations(
...
+        savedStateDataSource.savedState.take(1),
+        timelineRepository.homeTimelines(),
+    ) { savedState, homeTimelines ->
+        savedState to homeTimelines
+    }.mapToMutation { (savedState, homeTimelines) ->
+        val tabUri = currentTabUri
+            ?: savedState.signedInUserPreferences()
+                ?.lastViewedHomeTimelineUri
+                ?.takeIf { uri ->
+                    homeTimelines.any { it.isPinned && it.uri == uri }
```
tunjid, line 143: Nice improvement here with `List.any { }`

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
tunjid: Resolved.

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeScreen.kt
joelmuraguri: No that was not my intention. Fixing this

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeScreen.kt
joelmuraguri: Also this, fixing it

feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeViewModel.kt
joelmuraguri: Thanks for noticing that

tunjid (approved):

Issue #404, opened 2025-09-04
Draggable divider conflicts with gesture navigation


alexvanyo:
Right now, trying to drag the divider from the edges of the screen usually conflicts with the back gesture navigation.

Adding `Modifier.systemGestureExclusion` should allow dragging the divider without triggering the system back gesture.

tunjid:
@alexvanyo I tried the modifier, but it doesn't seem to work :(

I tried changing the Modifier order to before layout and sizing and it made no difference.

https://github.com/tunjid/heron/pull/423

PR #423, opened 2025-09-06
Try to exclude system gestures from pane anchor thumb boundaries

tunjid:
Fixes https://github.com/tunjid/heron/issues/404

### DIFF
--- a/scaffold/src/androidMain/kotlin/com/tunjid/heron/scaffold/scaffold/AppPanes.android.kt
+++ b/scaffold/src/androidMain/kotlin/com/tunjid/heron/scaffold/scaffold/AppPanes.android.kt
@@ -1,0 +1,31 @@
+/*
+ *    Copyright 2024 Adetunji Dahunsi
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/scaffold/AppPanes.kt
```diff
@@ -207,6 +207,7 @@ internal class PaneAnchorState {
                     }
                     .fillMaxHeight()
                     .width(DraggableDividerSizeDp)
+                    .platformSystemGestureExclusion()
```
alexvanyo: The `systemGestureExclusion()` can only exclude a limited portion of the height - where this is placed right now, it is trying to exclude the entire `fillMaxHeight()` `Box` and not just the thumb itself. If you move this to just above the `.width(thumbWidth)` applied to the `Surface` below then it should work (and you should be able to use the simple no-args `Modifier.systemGestureExclusion()` variant)

tunjid: This worked, ty!

scaffold/src/androidMain/kotlin/com/tunjid/heron/scaffold/scaffold/AppPanes.android.kt
```diff
@@ -0,0 +1,31 @@
...
+
+internal actual fun Modifier.platformSystemGestureExclusion(): Modifier =
+    systemGestureExclusion { layoutCoords ->
+        Rect(
+            left = 0f,
+            top = 0f,
+            right = layoutCoords.size.width.toFloat(),
+            bottom = layoutCoords.size.height.toFloat(),
+        )
+    }
```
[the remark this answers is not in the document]
tunjid: Fixed.

PR #452, opened 2025-09-11
Add Download Images Option Button

joelmuraguri:
### Changes Made

- Added Downloaded icon to the UI
- Added string resource for the button text description
- Introduced `Download` in `PostInteractionButton` sealed class

### Notes

- This PR only introduces the UI elements
- No download logic is implemented yet

### DIFF
--- a/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
+++ b/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
@@ -246,12 +246,12 @@ private inline fun PostInteractionsButtons(
                         PostInteractionButton.Comment -> onReplyToPost()
                         PostInteractionButton.Like -> onPostInteraction(
                             when (likeUri) {
-                                null -> com.tunjid.heron.data.core.models.Post.Interaction.Create.Like(
+                                null -> Post.Interaction.Create.Like(
                                     postId = postId,
                                     postUri = postUri,
                                 )
 
-                                else -> com.tunjid.heron.data.core.models.Post.Interaction.Delete.Unlike(
+                                else -> Post.Interaction.Delete.Unlike(
                                     postUri = postUri,
                                     likeUri = likeUri,
                                 )
@@ -260,12 +260,12 @@ private inline fun PostInteractionsButtons(
 
                         PostInteractionButton.Repost -> onPostInteraction(
                             when (repostUri) {
-                                null -> com.tunjid.heron.data.core.models.Post.Interaction.Create.Repost(
+                                null -> Post.Interaction.Create.Repost(
                                     postId = postId,
                                     postUri = postUri,
                                 )
 
-                                else -> com.tunjid.heron.data.core.models.Post.Interaction.Delete.RemoveRepost(
+                                else -> Post.Interaction.Delete.RemoveRepost(
                                     postUri = postUri,
                                     repostUri = repostUri,
                                 )
@@ -273,12 +273,12 @@ private inline fun PostInteractionsButtons(
                         )
                         PostInteractionButton.Bookmark -> onPostInteraction(
                             when (isBookmarked) {
-                                false -> com.tunjid.heron.data.core.models.Post.Interaction.Create.Bookmark(
+                                false -> Post.Interaction.Create.Bookmark(
                                     postId = postId,
                                     postUri = postUri,
                                 )
 
-                                true -> com.tunjid.heron.data.core.models.Post.Interaction.Delete.RemoveBookmark(
+                                true -> Post.Interaction.Delete.RemoveBookmark(
                                     postUri = postUri,
                                 )
                             },


tunjid:
For this, lets restrict the download button to only show on the `GalleryScreen`. We need the space on regular posts for the share button and an over flow to mute, report or block.

We should be able to do this my restricting changes to just the `MediaPostInteractions` Composable:

```kotlin
@Composable
fun MediaPostInteractions(
    post: Post,
    sharedElementPrefix: String,
    paneMovableElementSharedTransitionScope: MovableElementSharedTransitionScope,
    onReplyToPost: () -> Unit,
    onPostInteraction: (Post.Interaction) -> Unit,
) {
    Column(
        modifier = modifier,
        verticalArrangement = Arrangement.spacedBy(12.dp),
        horizontalAlignment = Alignment.CenterHorizontally,
    ) {
        PostInteractionsButtons(
            ...
        )
        // Mute  button can go here
    }
}
```

For the download button, let's put it to the right of the `MediaPoster` Composable inside the `videoPlayerController.MediaOverlay`.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
```diff
@@ -539,7 +555,7 @@ private val Timeline.Presentation.postInteractionArrangement: Arrangement.Horizo
         Timeline.Presentation.Media.Grid -> Arrangement.SpaceBetween
     }
 
-private val Timeline.Presentation.actionIconSize
+val Timeline.Presentation.actionIconSize
```
tunjid: Lets keep this as `private` too.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
```diff
@@ -73,6 +75,8 @@ import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.tunjid.composables.ui.animate
 import com.tunjid.heron.data.core.models.Post
+import com.tunjid.heron.data.core.models.Post.Interaction.Create
```
tunjid: Let's remove these two import statements:

```kotlin
import com.tunjid.heron.data.core.models.Post.Interaction.Create
import com.tunjid.heron.data.core.models.Post.Interaction.Delete
```

and use the fully qualified import like so: `com.tunjid.heron.data.core.models.Post.Interaction.Create` and `Post.Interaction.Create.Repost`

joelmuraguri: Why use this approach instead of imports?

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
joelmuraguri: Thanks for pointing this out. Already updated it

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
tunjid: I prefer fully qualified imports bc they are explicit. There are multiple classes names in the project with the name `Create`, `Message.Create` for example. I want to be explicit that this creates a `Post` interaction.

It can be inferred from the context, but being explicit is safer.

There is still an import, its the import for the `Post`:

```kotlin
import com.tunjid.heron.data.core.models.Post.Interaction.Create
```

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
```diff
@@ -232,12 +246,12 @@ private inline fun PostInteractionsButtons(
                         PostInteractionButton.Comment -> onReplyToPost()
                         PostInteractionButton.Like -> onPostInteraction(
                             when (likeUri) {
-                                null -> Post.Interaction.Create.Like(
+                                null -> com.tunjid.heron.data.core.models.Post.Interaction.Create.Like(
```
tunjid: Oh I don't mean the full import like this. I mean `Post.Interaction.Create.Like` instead of just `Create.Like`.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
```diff
@@ -246,29 +259,32 @@ private inline fun PostInteractionsButtons(
...
 
-                                true -> Post.Interaction.Delete.RemoveBookmark(
+                                true -> RemoveBookmark(
                                     postUri = postUri,
                                 )
                             },
                         )
+                        PostInteractionButton.Download -> {
+
+                        }
```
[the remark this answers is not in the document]
tunjid: Will be done in a follow up.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
joelmuraguri: Aah alright. That's clear now. I have made the needed changes.

PR #470, opened 2025-09-17
Add mute/unmute option toggle

joelmuraguri:
- Added a mute/unmute button to video playback.
- Tied button to VideoPlayerController.isMuted for toggling sound.
- Improves UX by letting users control audio during autoplay.

## Video Demo


tunjid:
Thanks, we need to rework this UI a bit as all the actions have the same priority, but some should have more priority than others. We'll fix this in a follow up.

PR #508, opened 2025-09-30
Write SavedState model polymorphic serialization tests

joelmuraguri:


tunjid:
Oh so this is testing the wrong thing. I was hoping to test the Versioned saved state migrations, which will implicitly test this as well.

A user migrating the app from `SavedStateVersion0` should be able to migrate to `VersionedSavedState` with no data loss. Same for all instances of `SavedStateVersion`. And since saved state is written with protobufs, we don't need to consider cbor for these tests.

As user feature requests come in, the `VersionedSavedState` will keep changing, and any migration path should still work.

It's probs best to close this PR and work on another.

PR #516, opened 2025-10-01
Make Profile bio links clickable

joelmuraguri:
**Fixes**: Links in profile bio were not clickable.

- Added link parsing for profile bio text (external links, hashtags, user mentions).
- Implemented `ProfileBio` composable with tap detection for annotated links.
- Navigation now routes to the correct destination or opens external URLs.
- Bio text uses theme-aware colors; links are styled in blue.

### DIFF
--- a/feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
+++ b/feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
@@ -23,7 +23,6 @@ import androidx.compose.animation.core.animateDpAsState
 import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
-import androidx.compose.foundation.gestures.detectTapGestures
 import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
@@ -46,7 +45,6 @@ import androidx.compose.foundation.pager.PagerState
 import androidx.compose.foundation.pager.rememberPagerState
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.foundation.text.BasicText
 import androidx.compose.material3.CircularWavyProgressIndicator
 import androidx.compose.material3.ExperimentalMaterial3ExpressiveApi
 import androidx.compose.material3.LocalTextStyle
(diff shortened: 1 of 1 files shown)


feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
```diff
@@ -520,6 +536,45 @@ private fun ProfileHeader(
...
+
+    val annotatedText = rememberFormattedTextPost(
+        text = bio,
+        textLinks = textLinks,
+        onLinkTargetClicked = onLinkTargetClicked,
+    )
+
+    var textLayoutResult by remember { mutableStateOf<TextLayoutResult?>(null) }
+
+    BasicText(
```
tunjid: Do we need to do custom click handling for the text? 

`rememberFormattedTextPost` already returns an `AnnotatedString` with clickable spans. We should just be able to use regular Material3 `Text` here, and the click handling should work out of the box.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
```diff
@@ -274,6 +282,9 @@ internal fun ProfileScreen(
                         ),
                     )
                 },
+                onLinkTargetClicked = { target ->
+                    actions(Action.BioLinkClicked(target))
```
tunjid: In this file, we already handle link target clicks. We should be able to reuse the existing action: https://github.com/tunjid/heron/blob/main/feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt#L931

tunjid (commented):
Some changes suggested.

Consider adding the following method to `ui:core`, and using it in `ProfileScreen` for both profile bio links and other places where `LinkTarget.Navigable` is used.

```kotlin
@Composable
fun navigableLinkTargetHandler(
    onNavigableLinkTargetClicked: (LinkTarget.Navigable) -> Unit
): (LinkTarget) -> Unit {
    val uriHandler = LocalUriHandler.current
    return remember(onNavigableLinkTargetClicked) {
        { linkTarget ->
            when(linkTarget) {
                is LinkTarget.ExternalLink -> runCatching {
                    uriHandler.openUri(linkTarget.uri.uri)
                }
                is LinkTarget.Navigable -> onNavigableLinkTargetClicked(
                    linkTarget
                )
            }
        }
    }
}
```

We should also do a search in the code base for usages of `if (linkTarget is LinkTarget.Navigable)`, and update them to use this method.

That's probably best for a separate PR however.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
joelmuraguri: Thnaks for the clarification. This is updated

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
joelmuraguri: Again thanks for this!

joelmuraguri:
@tunjid

PR #530, opened 2025-10-07
Allow uploading of profile pictures; avatars and  headers

joelmuraguri:
## Demo

https://github.com/user-attachments/assets/4927f6c9-c311-44a0-a04b-2036934750e3


scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/navigation/NavigationMutator.kt
```diff
@@ -150,6 +150,14 @@ fun conversationDestination(
     referringRouteOption = referringRouteOption,
 )
 
+fun editProfileDestination(
+    profile: Profile,
+): NavigationAction.Destination = pathDestination(
+    path = "/editProfile/${profile.did.id}",
```
tunjid: Let's make this path `"/profile/${profile.did.id}/edit"` instead.

feature/edit-profile/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+                implementation(project(":data:core"))
+                implementation(project(":scaffold"))
+                implementation(project(":feature:template"))
+
+                implementation(libs.kotlinx.coroutines.core)
+                implementation(libs.kotlinx.datetime)
+                implementation(libs.kotlinx.serialization.cbor)
+                implementation(libs.kotlinx.serialization.json)
+
+                implementation(libs.filekit.coil)
```
tunjid: In the Compose post screen, we don't have direct dependencies on filekit, instead we use the wrappers in the `ui:media` module. We should do the same here too.

feature/edit-profile/src/commonMain/kotlin/com/tunjid/heron/editprofile/State.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.tunjid.treenav.strings.Route
+import io.github.vinceglb.filekit.PlatformFile
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.Transient
+
+@Serializable
+data class State(
+    val profile: Profile,
+    @Transient
+    val updatedAvatar: PlatformFile? = null,
```
tunjid: We don't have any control over the `PlaformFile` type, and it doesn't have any guarantees on if it is `@Stable`, or implements `equals()` or `hashcode()`, so it's not a good candidate to be in the `State` class. We should replace it with the [`MediaItem`](https://github.com/tunjid/heron/blob/365cdecb8771374c39b8a327c89f6e70d70cbba2/ui/media/src/commonMain/kotlin/com/tunjid/heron/media/picker/MediaItem.kt#L26) type we control.

feature/edit-profile/src/commonMain/kotlin/com/tunjid/heron/editprofile/EditProfileScreen.kt
```diff
@@ -0,0 +1,247 @@
...
+import org.jetbrains.compose.resources.stringResource
+
+@Composable
+internal fun EditProfileScreen(
+    paneScaffoldState: PaneScaffoldState,
+    state: State,
+    actions: (Action) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val avatarPicker = rememberFilePickerLauncher(
```
tunjid: Using file kit APIs directly opens us up to changes if filekit makes any changes to their API, so lets not leak its API in the feature module. Instead, we should use the wrapper with our own API. If file kit changes their API, we only need to update our wrapper in one place.

An example of using the wrapper API can be seen [here](https://github.com/tunjid/heron/blob/365cdecb8771374c39b8a327c89f6e70d70cbba2/feature/compose/src/commonMain/kotlin/com/tunjid/heron/compose/ui/ComposePostBottomBar.kt#L69).

tunjid: It also returns types from the `ui:media` module that are safe to put inside the `State` class for the UI.

tunjid (commented):
Great first pass, the changes requested are to use APIs and types we fully own, instead of using the filekit APIs directly.

This consolidates file management and media API usages to APIs we fully control, and we can change to backing API to something else when we like instead of updating the whole app.

joelmuraguri:
> Great first pass, the changes requested are to use APIs and types we fully own, instead of using the filekit APIs directly.
> 
> This consolidates file management and media API usages to APIs we fully control, and we can change to backing API to something else when we like instead of updating the whole app.

@tunjid Thanks for pointing this out and the why.

composeApp/src/commonMain/kotlin/com/tunjid/heron/di/AppGraph.kt
```diff
@@ -65,6 +66,7 @@ interface AppGraph {
             @Includes composeBindings: ComposeBindings,
             @Includes conversationBindings: ConversationBindings,
             @Includes feedBindings: FeedBindings,
+            @Includes editProfileBindings: EditProfileBindings,
```
tunjid, line 68: Minor: Should e alphabetical, `e` before `f`

tunjid (approved):
One minor change requested.

PR #532, opened 2025-10-09
Implement uploading profile images

joelmuraguri:


data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
```diff
@@ -592,6 +603,55 @@ internal class OfflineProfileRepository @Inject constructor(
...
+                        description = update.bio,
+                        avatar = avatarBlob ?: existingProfile.avatar,
+                        banner = bannerBlob ?: existingProfile.banner,
+                    )
+                        .asJsonContent(BskyProfile.serializer()),
+                )
+
+                networkService.runCatchingWithMonitoredNetworkRetry {
+                    putRecord(request)
+                }
```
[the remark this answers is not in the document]
tunjid: The AI is right about this, we should apply its suggestion.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/Profile.kt
```diff
@@ -69,6 +69,15 @@ data class Profile(
         val createdFeedGeneratorCount: Long,
         val createdStarterPackCount: Long,
     )
+
+    @Serializable
+    data class Update(
+        val profileId: ProfileId,
+        val displayName: String,
+        val bio: String,
```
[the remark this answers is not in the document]
tunjid: This is fine.

feature/edit-profile/src/commonMain/kotlin/com/tunjid/heron/editprofile/EditProfileViewModel.kt
```diff
@@ -80,3 +97,44 @@ private fun Flow<Action.BannerPicked>.bannerPickedMutations(): Flow<Mutation<Sta
     mapToMutation {
         copy(updatedBanner = it.item)
     }
+
+private fun Flow<Action.SaveProfile>.saveProfileMutations(
+    navActions: (NavigationMutation) -> Unit,
+    writeQueue: WriteQueue,
+): Flow<Mutation<State>> =
+    mapToManyMutations { action ->
```
tunjid: To be safe, we should use `mapLatestToManyMutations` instead, this will prevent having two update running at the same time. Any new saves will cancel the last request to enqueue. However, if its already queued it won't make a difference.

feature/edit-profile/src/commonMain/kotlin/com/tunjid/heron/editprofile/EditProfileViewModel.kt
```diff
@@ -80,3 +97,44 @@ private fun Flow<Action.BannerPicked>.bannerPickedMutations(): Flow<Mutation<Sta
...
+
+        val updateWrite = withContext(Dispatchers.IO) {
```
tunjid: Using `Dispatchers.IO` here doesn't really do anything, all it does is create an object on an IO thread. It should be removed.

feature/edit-profile/src/commonMain/kotlin/com/tunjid/heron/editprofile/EditProfileViewModel.kt
```diff
@@ -80,3 +97,44 @@ private fun Flow<Action.BannerPicked>.bannerPickedMutations(): Flow<Mutation<Sta
...
+        }
+
+        writeQueue.enqueue(updateWrite)
+        writeQueue.awaitDequeue(updateWrite)
+
+        emit {
+            copy(
+                messages = messages + ScaffoldMessage.Text("Profile updated successfully!"),
+            )
+        }
```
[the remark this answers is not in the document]
tunjid: This is true, submitting to the queue is designed to be fire and forget, the outcome is not supposed to be reported to the user.

There will be a screen in the future to see failed writes and other things, but for now, just submit it to the queue, a message doesn't need to be shown.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
tunjid: After we put the record, we should also update the signed in user so changes to the profile are seen.

joelmuraguri:
@tunjid

feature/edit-profile/src/commonMain/kotlin/com/tunjid/heron/editprofile/EditProfileViewModel.kt
```diff
@@ -80,3 +98,37 @@ private fun Flow<Action.BannerPicked>.bannerPickedMutations(): Flow<Mutation<Sta
...
+        writeQueue.awaitDequeue(updateWrite)
+
+        emitAll(
+            flowOf(Action.Navigate.Pop).consumeNavigationActions(
+                navigationMutationConsumer = navActions,
+            ),
+        )
+    }
+
+private suspend fun MediaItem.Photo.toMediaFile(): MediaFile.Photo {
```
tunjid, line 128: Oh I'm sorry, I made a mistake earlier. I see why you used `Dispatchers.IO`, it was for `readBytes()`. It is in fact necessary.

tunjid:
Actually, another correction, `readBytes` suspends properly, so there is no need to use `Dispatcher.IO`. Merging.

Issue #526, opened 2025-10-06
Support embedded records in posts


tunjid:
**Is your feature request related to a problem? Please describe.**
Posts may currently have a feed,  list, starter pack, or other record embedded. They are not currently displayed.

**Describe the solution you'd like**
These records should be displayed.

**Describe alternatives you've considered**

**Additional context**
This task will require adding a new field to the `PostEntity`, then adding a migration for it. `PostEntity` needs to have a reference to it's embedded `RecordUri` if it has one. 

** Step 1

Update `PostEntity.RecordData` to:

```kotlin
data class PostEntity(
    @PrimaryKey
    val uri: PostUri,
    ...
    @Embedded
    val record: RecordData?,
) {
    data class RecordData(
        val text: String,
        val base64EncodedRecord: String?,
        val embeddedRecordUri: RecordUri?,
        val createdAt: Instant,
    )
}
```

** Step 2

Add an entry in `UriConverters` for `RecordId`.

** Step 3

Add an `AutoMigration` for this change, and bump the current database version by 1

```kotlin
@Database(
    version = /** existing version + 1**/,
    entities = [
        ...
    ],
    autoMigrations = [
        ...
        // Migration 23 - 24 is a manual migration
        // Add embeddedRecordUri to PostEntity
        AutoMigration(from = existingVersion, to = existingVersion + 1)
    ],
    exportSchema = true,
)
@TypeConverters(
    DateConverters::class,
    UriConverters::class,
    IdConverters::class,
)
@ConstructedBy(AppDatabaseConstructor::class)
abstract class AppDatabase : RoomDatabase() {
    ...
}
```

Subsequent PRs will actually add this data to posts in the database and retrieve them in post queries.

tunjid:
Some updates in https://github.com/tunjid/heron/pull/597

Issue #549, opened 2025-10-13
Support labelers and label persistence


tunjid:
**Is your feature request related to a problem? Please describe.**
The app currently has public model definitions for labels and labelers, but has no support for persisting labelers or labels to the sqlite database.

**Describe the solution you'd like**
Entities for these types should be added so they can be persisted and retrieved for profiles and posts.

**Details**
Read this [doc](https://docs.bsky.app/docs/advanced-guides/moderation#labelers) to understand labels and labelers.

Labelers are really just profiles with metadata that say they are actually labelers. This metadata is stored in `ProfileEntity.Associated`. For this issue, we need to save the definitions for each description in a `Labeler`.

* [ ] Add `LabelDefinitionEntity`
* [ ] Add Fetch labelers for the signed in user, and persist them and their label definitions.
* [ ] Implement `labelers` in `TimelineRepository`.

Add a `LabelDefinitionEntity`:

```kotlin
@Entity(
    tableName = "labelDefinitions",
    primaryKeys = ["labelerId", "identifier"],
    foreignKeys = [
        ForeignKey(
            entity = ProfileEntity::class,
            parentColumns = ["did"],
            childColumns = ["labelerId"],
            onDelete = ForeignKey.CASCADE,
        ),
    ],
    indices = [
        Index(value = ["creatorId"]),
        Index(value = ["identifier"]),
    ],
)
data class LabelDefinitionEntity(
        val creatorId: ProfileId,
        val identifier: String,
        val adultOnly: Boolean,
        val blurs: String,
        val defaultSetting: String,
        val severity: String,
        @ColumnInfo(typeAffinity = ColumnInfo.BLOB)
        val localeInfo: ByteArray,
    )
```

Not that the definition:
* Has compound primary keys
* Stores the `localeInfo` as a blob.

From the doc linked above, a `Labeler` describes itself as:

```kotlin
{
  "$type": "app.bsky.labeler.service",
  "policies": {
    "labelValues": ["porn", "spider"],
    "labelValueDefinitions": [
      {
        "identifier": "spider",
        "severity": "alert",
        "blurs": "media",
        "defaultSetting": "warn",
        "locales": [
          {"lang": "en", "name": "Spider Warning", "description": "Spider!!!"}
        ]
      }
    ]
  },
  "subjectTypes": ["record"],
  "subjectCollections": ["app.bsky.feed.post", "app.bsky.actor.profile"],
[...]

PR #573, opened 2025-10-26
Add support of embedded records in posts

joelmuraguri:
Fixes https://github.com/tunjid/heron/issues/526

### DIFF
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/network/models/TimelineItemConversions.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/network/models/TimelineItemConversions.kt
@@ -25,7 +25,6 @@ import app.bsky.feed.ReplyRefRootUnion
 import com.tunjid.heron.data.core.models.Constants
 import com.tunjid.heron.data.core.types.FeedGeneratorUri
 import com.tunjid.heron.data.core.types.ImageUri
-import com.tunjid.heron.data.core.types.LabelerUri
 import com.tunjid.heron.data.core.types.ListUri
 import com.tunjid.heron.data.core.types.PostUri
 import com.tunjid.heron.data.core.types.ProfileHandle
@@ -78,7 +77,7 @@ internal fun FeedViewPost.feedItemEntity(
             is RecordViewRecordUnion.FeedGeneratorView -> FeedGeneratorUri(recordUnion.value.uri.atUri)
             is RecordViewRecordUnion.GraphListView -> ListUri(recordUnion.value.uri.atUri)
             is RecordViewRecordUnion.GraphStarterPackViewBasic -> StarterPackUri(recordUnion.value.uri.atUri)
-            is RecordViewRecordUnion.LabelerLabelerView -> LabelerUri(recordUnion.value.uri.atUri)
+            is RecordViewRecordUnion.LabelerLabelerView -> null
             is RecordViewRecordUnion.Unknown -> null
             is RecordViewRecordUnion.ViewBlocked -> null // TODO: Handle blocked/muted posts later
             is RecordViewRecordUnion.ViewDetached -> null // TODO: Handle detached posts
@@ -90,7 +89,7 @@ internal fun FeedViewPost.feedItemEntity(
             is RecordViewRecordUnion.FeedGeneratorView -> FeedGeneratorUri(recordUnion.value.uri.atUri)
             is RecordViewRecordUnion.GraphListView -> ListUri(recordUnion.value.uri.atUri)
             is RecordViewRecordUnion.GraphStarterPackViewBasic -> StarterPackUri(recordUnion.value.uri.atUri)
-            is RecordViewRecordUnion.LabelerLabelerView -> LabelerUri(recordUnion.value.uri.atUri)
+            is RecordViewRecordUnion.LabelerLabelerView -> null
             is RecordViewRecordUnion.Unknown -> null
             is RecordViewRecordUnion.ViewBlocked -> null // TODO
             is RecordViewRecordUnion.ViewDetached -> null // TODO
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/RecordRepository.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/RecordRepository.kt
@@ -22,7 +22,6 @@ import app.bsky.graph.GetListQueryParams
 import app.bsky.graph.GetStarterPackQueryParams
 import com.tunjid.heron.data.core.models.Record
 import com.tunjid.heron.data.core.types.FeedGeneratorUri
-import com.tunjid.heron.data.core.types.LabelerUri
 import com.tunjid.heron.data.core.types.ListUri
 import com.tunjid.heron.data.core.types.PostUri
 import com.tunjid.heron.data.core.types.RecordUri
@@ -85,7 +84,6 @@ internal class OfflineRecordRepository @Inject constructor(
                         )
                     }
                     .map { it.firstOrNull()?.asExternalModel(quote = null) }
-            else -> emptyFlow() // safely ignore LabelerUri and any unknown ones for now
         }
             .filterNotNull()
             .withRefresh {
@@ -160,6 +158,5 @@ internal suspend fun NetworkService.refresh(
             ?.let { starterPackView ->
                 multipleEntitySaverProvider.saveInTransaction { add(starterPackView) }
             }
-        is LabelerUri -> null // not supported yet
     }
 }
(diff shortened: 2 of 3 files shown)


data/core/src/commonMain/kotlin/com/tunjid/heron/data/network/models/PostConversions.kt
```diff
@@ -273,18 +278,38 @@ internal fun ReplyRefParentUnion.postViewerStatisticsEntity(
...
             text = bskyPost.text,
             base64EncodedRecord = bskyPost.toPostRecord().toUrlEncodedBase64(),
             createdAt = bskyPost.createdAt,
+            embeddedRecordUri = embeddedUri,
         )
     } catch (_: Exception) {
         null
     }
 
+private fun String.toRecordUriOrNull(): RecordUri? =
```
tunjid: This method has duplicates. In `Collections.kt` theres `Uri.asRecordUri()`, lets add a internal overload for `String.asRecordUri()` that uses its logic, then delegate to it for `Uri.asRecordUri()`. 

That `String` version should then be used for the database converter too.

This also has the benefit of using the constants already defined in `Collections.kt`.

tunjid:
In creating the follow up issue: https://github.com/tunjid/heron/issues/574 I noticed a new field also needs to be added to `TimelineItemEntity`. We also need to add a `embeddedRecordUri` to it so when fetching posts for a timeline, we can efficient look up the full embedded record too. The only file that creates a `TimelineItemEntity` is `TimelineItemConversions.kt` You should have the following:

```kotlin
data class TimelineItemEntity(
    val postUri: PostUri,
    val viewingProfileId: ProfileId?,
    val sourceId: String,
    val embeddedRecordId: RecordId?
    @Embedded
    val reply: FeedReplyEntity?,
    val reposter: ProfileId?,
    @ColumnInfo(
        defaultValue = "false",
    )
    val hasMedia: Boolean,
    val isPinned: Boolean,
    val indexedAt: Instant,
    // Timeline items are unique to the profile viewing them, and these other fields
    @PrimaryKey
    val id: String = "${viewingProfileId?.id}-$sourceId-${postUri.uri}-${reposter?.id}",
)
```

This should be done as part of this PR, using the same `AutoMigration`. You will need to t clear the app data on your device to reuse the migration, as the migration has already occurred when you are testing.

To verify everything works as it should:

* Uninstall the app.
* Build and install the app on the main branch.
* Switch to this branch and build and install the app.

If no crashes, the migration is successful.

joelmuraguri:
@tunjid

data/core/src/commonMain/kotlin/com/tunjid/heron/data/network/models/TimelineItemConversions.kt
```diff
@@ -67,6 +69,24 @@ internal fun FeedViewPost.feedItemEntity(
         null,
         -> post.indexedAt
     },
+    embeddedRecordId = when (val embed = post.embed) {
```
tunjid: This was a typo from me 😞. Should be `embeddedRecordUri`, not `embededRecordId`. There'sno need to create a new `RecordId` type.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/network/models/TimelineItemConversions.kt
```diff
@@ -67,6 +69,24 @@ internal fun FeedViewPost.feedItemEntity(
...
+        is PostViewEmbedUnion.RecordView -> when (val recordUnion = embed.value.record) {
+            is RecordViewRecordUnion.ViewRecord -> RecordId(recordUnion.value.uri.atUri)
```
tunjid: We want to make sure these Uris are actually for records like posts, starter packs, feeds or lists. So we should be exhaustive with these checks. All types should be accounted for in the `when` so if a new type is introduced, we get a compilation error instead of just returning null. 

The same applies to the `RecordViewRecordUnion` types. For `RecordViewRecordUnion.ViewRecord`, return a `PostUri`. For `ViewBlocked`, `ViewDetached` and `ViewNotFound` return `null` for now. There's a an issue to deal with blocked or muted posts more comprehensively later. Feel free to add a TODO comment.

```
when(val embed = post.embed) {
        // Return explicit record Uri types like `FeedGeneratorUri` and so on, else null
        is PostViewEmbedUnion.RecordWithMediaView -> when(embed.value.record) {
            is RecordViewRecordUnion.FeedGeneratorView -> TODO()
            is RecordViewRecordUnion.GraphListView -> TODO()
            is RecordViewRecordUnion.GraphStarterPackViewBasic -> TODO()
            is RecordViewRecordUnion.LabelerLabelerView -> TODO()
            is RecordViewRecordUnion.Unknown -> TODO()
            is RecordViewRecordUnion.ViewBlocked -> TODO()
            is RecordViewRecordUnion.ViewDetached -> TODO()
            is RecordViewRecordUnion.ViewNotFound -> TODO()
            is RecordViewRecordUnion.ViewRecord -> TODO()
        }
                // Return explicit record Uri types like `FeedGeneratorUri` and so on, else null
        is PostViewEmbedUnion.RecordView -> when(embed.value.record.record) {
            is RecordViewRecordUnion.FeedGeneratorView -> TODO()
            is RecordViewRecordUnion.GraphListView -> TODO()
            is RecordViewRecordUnion.GraphStarterPackViewBasic -> TODO()
            is RecordViewRecordUnion.LabelerLabelerView -> TODO()
            is RecordViewRecordUnion.Unknown -> TODO()
            is RecordViewRecordUnion.ViewBlocked -> TODO()
            is RecordViewRecordUnion.ViewDetached -> TODO()
            is RecordViewRecordUnion.ViewNotFound -> TODO()
            is RecordViewRecordUnion.ViewRecord -> TODO()
        }
        is PostViewEmbedUnion.ExternalView,
        is PostViewEmbedUnion.ImagesView,
        is PostViewEmbedUnion.Unknown,
        is PostViewEmbedUnion.VideoView,
        null -> null
    }
```

joelmuraguri:
@tunjid

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/RecordRepository.kt
```diff
@@ -157,5 +159,6 @@ internal suspend fun NetworkService.refresh(
             ?.let { starterPackView ->
                 multipleEntitySaverProvider.saveInTransaction { add(starterPackView) }
             }
+        is LabelerUri -> TODO()
```
tunjid: This will crash the app. It's better to not add `LabelerUri` in this PR, and add it in a follow up PR when we're ready to support it.

joelmuraguri: Since when needs all its branches to be exhausted. I will return an empty flow and null for now.

tunjid: Won't it be safer to just return null for `RecordViewRecordUnion.LabelerLabelerView`? The app doesn't know what to do with it yet. This issue is the place to actually add label support: https://github.com/tunjid/heron/issues/549

tunjid: Adding it here would just be scope creep. This PR and its issue has a well defined scope, don't let future scoped work creep into it.

joelmuraguri: Yes making `RecordViewRecordUnion.LabelerLabelerView` null is the correct way. Let me update the code

data/core/src/commonMain/kotlin/com/tunjid/heron/data/network/models/TimelineItemConversions.kt
```diff
@@ -67,6 +71,38 @@ internal fun FeedViewPost.feedItemEntity(
         null,
         -> post.indexedAt
     },
+    embeddedRecordUri = when (val embed = post.embed) {
```
tunjid, line 74: One last note, this method of extracting a `RecordUri` is significantly more efficient that the string matching used in

```
   try {
        val bskyPost = decodeAs<BskyPost>()

        val embeddedUri: RecordUri? = when (val embed = bskyPost.embed) {
            is PostEmbedUnion.Record ->
                embed.value.record.uri.atUri.asRecordUri()

            is PostEmbedUnion.RecordWithMedia ->
                embed.value.record.record.uri.atUri.asRecordUri()

            else -> null
        }
```

Perhaps we should extract 

```kotlin
fun PostViewEmbedUnion.RecordView.embeddedRecordUri(): RecordUri?
```
 as an internal method and use it in both places?

The database can continue using the sting match. Thanks for being so receptive to feedback.

tunjid: Ah never mind, different types.

joelmuraguri: Got it! Thanks for the clarification and the earlier suggestion.

PR #583, opened 2025-10-28
Fetch embedded records in posts

### LINKED ISSUE #574: Fetch embedded records in posts
Posts may currently have a feed,  list, starter pack, or other record embedded. They are not currently displayed.

These records should be displayed.

**Additional context**
After completion of https://github.com/tunjid/heron/issues/526, these embedded records will need to be fetched.

** Step 1:

The `Post` class currently has a `quote` field. This field should be deprecated, but not removed as it is serialized. a new field called `embeddedRecord` with type `Record?` should be added (it should be the last field in the `Post` class).

Instances of a `Post` are largely created with the `post()` method in `PostConversions.kt`. This method currently takes a `quote` argument for an embedded `Post` record. Update it, and its usages, to take the `embeddedRecord` argument instead.

** Step 2

In `TimelineRepository`, in both `observeTimeline` and `postThreadedItems` use the following method:

```kotlin
postDao.embeddedPosts(
    viewingProfileId = signedInProfileId?.id,
    postUris = postUris,
 )
```
A method will need to be added to check the `embeddedRecordUri` for the post or feed item to then fetch the embedded records.

joelmuraguri:
Fixes #574

### DIFF
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/TimelineRepository.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/TimelineRepository.kt
@@ -102,6 +102,7 @@ import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.firstOrNull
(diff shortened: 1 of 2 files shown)


joelmuraguri:
@tunjid  I was just about to commit some minor changes on this branch. How do I go about it? Or should I just add the changes in my next PR or create another separate PR for the minor changes

tunjid:
@joelmuraguri create a small PR with just those changes.

PR #609, opened 2025-11-07
Fetch and persist Labelers and LabelDefinitions

joelmuraguri:


data/core/src/commonMain/kotlin/com/tunjid/heron/data/di/Bindings.kt
```diff
@@ -184,6 +186,18 @@ class DataBindings(
         database: AppDatabase,
     ): LabelDao = database.labelDao()
 
+    @SingleIn(AppScope::class)
+    @Provides
```
tunjid: We can reuse the `LabelDao` for all label related database interactions. The `LabelDefinitionEntity` and `LabelerEntity` methods should be added to the `LabelDao`.

For example the `PostDao` works with `PostEntity`, `PostPostEntity`, `PostVideoEntity` and so on.

Having one DAO per `Entity` isn't a bad thing, but there isn't much value add in this case.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/multipleEntitysaver/SaveLabeler.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import app.bsky.labeler.LabelerViewDetailed
+import com.tunjid.heron.data.core.types.GenericUri
+import com.tunjid.heron.data.core.types.ProfileId
+import com.tunjid.heron.data.database.entities.LabelerEntity
+
+internal fun MultipleEntitySaver.add(
+    labeler: LabelerViewDetailed,
+) {
+    val creator = labeler.creator
```
tunjid, line 12: You should also be calling `MultipleEntityProvider.add(viewingProfileId, creator)` here from [`SaveProfileView.kt`](https://github.com/tunjid/heron/blob/93d5ad796cc35208516697e18c58415aaacede01/data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/multipleEntitysaver/SaveProfileView.kt#L4).

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/multipleEntitysaver/MultipleEntitySaver.kt
```diff
@@ -231,6 +243,8 @@ internal class MultipleEntitySaver(
         feedGeneratorDao.upsertFeedGenerators(feedGeneratorEntities.list)
 
         labelDao.upsertLabels(labelEntities.list)
+        labelDefinitionDao.upsertLabelValueDefinitions(labelDefinitionsEntities.list)
+        labelerDao.upsertLabelers(labelerEntities.list)
```
tunjid: I think the order of this should be:

* `labelers`
* `labelDefinitions`
* `labels`.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/multipleEntitysaver/SaveLabelValueDefinitions.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    val localeCbor = localeList.toUrlEncodedBase64()
+
+    add(
+        LabelDefinitionEntity(
+            creatorId = creatorId,
+            identifier = labelValueDefinition.identifier,
+            adultOnly = labelValueDefinition.adultOnly ?: false,
+            blurs = labelValueDefinition.blurs.value.lowercase(),
+            defaultSetting = labelValueDefinition.defaultSetting?.value?.lowercase() ?: "warn",
```
tunjid: I think we have warn as a constant somewhere, lets reuse that.

tunjid: `Label.Visibility.Warn.value`

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/LabelDefinitionEntity.kt
```diff
@@ -30,3 +33,12 @@ data class LabelDefinitionEntity(
     val severity: String,
     val localeInfoCbor: String,
 )
+
+fun LabelDefinitionEntity.asExternalModelWithLocale(): Pair<Label.Definition, Labeler.LocaleInfoList> =
```
tunjid: We don't seem to use the value of `Pair.second` when this method is called.

I think we should just include the `LocaleList` in the `Label.Definition` class. Foe the default one we use, it can just be an empty LocaleList. he next PR will will be replacing it anyway.

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/LabelDefinitionEntity.kt
```diff
@@ -30,3 +33,13 @@ data class LabelDefinitionEntity(
     val severity: String,
     val localeInfoCbor: String,
 )
+
+fun LabelDefinitionEntity.asExternalModel(): Label.Definition =
+    Label.Definition(
+        adultOnly = adultOnly,
+        blurs = Label.BlurTarget.valueOf(blurs.lowercase()),
```
tunjid: These may throw if a new enum is added by bluesky. For these enum lookups,
the following is safer:

```kotlin
Label.BlurTarget.entries.firstOrNull { it.name == name } ?: Label.BlurTarget...
```

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/daos/LabelerDao.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import androidx.room.Dao
+import androidx.room.Upsert
+import com.tunjid.heron.data.database.entities.LabelerEntity
+
+@Dao
+interface LabelerDao {
+    @Upsert
+    suspend fun upsertLabelers(
+        entities: List<LabelerEntity>,
```
tunjid: Let's fold this into the `LabelDao` too. Centralizes everything, we will split in the future if needed.

PR #610, opened 2025-11-07
Support labeler and label persistence

tunjid:


tunjid:
Recreating this with a rebase.

PR #646, opened 2025-11-14
Wire PostRepository quotes method in Posts module

joelmuraguri:


feature/posts/src/commonMain/kotlin/com/tunjid/heron/posts/PostsViewModel.kt
```diff
@@ -83,22 +100,48 @@ class ActualPostsViewModel(
...
-            // TODO: Fetch Saved posts
-        }
-        is PostsRequest.Quotes -> {
-            // TODO: Fetch quotes
+): Flow<Mutation<State>> = when (request) {
+    is PostsRequest.Saved -> {
+        // TODO: Fetch Saved posts - return empty for now
+        flowOf(mutationOf { copy(posts = emptyTiledList()) })
+    }
+    is PostsRequest.Quotes -> {
```
tunjid: This wont quite work. These should be paginated, and there's no pagination going on at the moment.  Also `PostRepository.quotes` should be updated to return `Flow<CursorList<TimelineItem>>` and not `Flow<CursorList<Post>>`

tunjid: For pagination, you will need to use `tilingMutations`. Examples include:

* [Timeline](https://github.com/tunjid/heron/blob/main/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/state/Timeline.kt#L114)
* [Search](https://github.com/tunjid/heron/blob/main/feature/search/src/commonMain/kotlin/com/tunjid/heron/search/SearchViewModel.kt#L475)

joelmuraguri: Noted, I'll have to do reading on this

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/PostRepository.kt
```diff
@@ -146,6 +149,7 @@ internal class OfflinePostRepository @Inject constructor(
     private val transactionWriter: TransactionWriter,
     private val fileManager: FileManager,
     private val savedStateDataSource: SavedStateDataSource,
+    private val timelineRepository: TimelineRepository,
```
tunjid: Repositories depending on repositories can lead to circular dependencies. Its safer to depend on the labelDao instead, and use it to read the labels. Its fine if we don't refresh the labels from the network here.

tunjid:
There are a few things about this that ought to be better, but I'll add so you can work on the UI.

1. The labelers are empty.
2. Ideally when consuming flows and flatmapping, we want to keep the number of database queries to a minimum.
    * `withPostEntity` currently hits the database twice. I'll replace it with a method to simply resolve the `PostUri`, no need to fetch the entire entity.
    * The `labelers` `Flow` can be pulled up to a utility class and shared by both `TimelineRepository`, and `PostRepository`. This is the `RecordResolver` interface. Some common methods have been pulled in. Filtering posts be label preferences can be common too, but that's for a separate PR.

PR #738, opened 2025-12-18
Moderation feature muted words implementation

joelmuraguri:


data/core/src/commonMain/kotlin/com/tunjid/heron/data/di/Bindings.kt
```diff
@@ -246,6 +249,12 @@ class DataBindings(
         database: AppDatabase,
     ): MessageDao = database.messagesDao()
 
+    @SingleIn(AppScope::class)
+    @Provides
+    fun providePreferencesDao(
```
tunjid: Oh I don't think we should persist this with SQL, unless there's a good reason to. It's easier to put this in the proto in `SavedStateDataSource`.

* Each user will have their own muted words and there's a preety good convention for that in the proto. With SQL atm its adhoc.
* The data is fundamentally unstructured.
* The data isn't going to be joined, or queried with any SQL like syntax.
* It only has one relation - the querying user.

Is there a concern you had that made you go for SQL?

joelmuraguri: Great point. I was thinking about SQL because I imagined the moderation sheet might need to handle multiple complex data types (muted words, blocked/muted accounts, labelers) and thought SQL would provide better organization. But you're absolutely right that:

- The data is unstructured where each preference type is independent
- We're already using proto successfully for all other user preferences
- SQL would be overkill for what's essentially simple key-value storage per user

Already making the changes

joelmuraguri:
@tunjid

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
```diff
@@ -0,0 +1,140 @@
...
+import com.tunjid.heron.data.core.utilities.Outcome
+import com.tunjid.heron.data.network.NetworkService
+import com.tunjid.heron.data.utilities.preferenceupdater.PreferenceUpdater
+import dev.zacsweers.metro.Inject
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+
+interface UserDataRepository {
+    fun mutedWords(): Flow<List<MutedWordPreference>>
```
tunjid: As this doesn't take any arguments, its better as a val.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/moderation/ModerationState.kt
```diff
@@ -0,0 +1,29 @@
...
+import androidx.compose.runtime.Stable
+import androidx.compose.ui.graphics.vector.ImageVector
+import com.tunjid.heron.timeline.ui.sheets.MutedWordsStateHolder
+
+sealed class ModerationOption(
+    val title: String,
+    val icon: ImageVector,
+) {
+    data object MuteWords : ModerationOption(
+        title = "Mute words and tags",
```
tunjid: Shouldn't this be a string resource?

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/moderation/ModerationState.kt
```diff
@@ -0,0 +1,29 @@
...
+        icon = Icons.Default.FilterAlt,
+    )
+
+    data object BlockUser : ModerationOption(
+        title = "Block user",
```
tunjid: Should also be a string resource.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    suspend fun createMutedWord(preference: MutedWordPreference): Outcome
```
tunjid: Each of these suspend methods return an outcome, which are not good for user ergonomics. Each time a user edits muted words, the update is enqueued, and completed at some point in the future.

For something like editing muted words, feedback in the UI should be immediate. I think all these add, remove and other methods should operate on data in memory, in the state holder. Then, when the user hits save, a single update data class with what the new muted words should be is sent.

This would also remove the code duplication with the preference updater u have for all the methods.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    suspend fun removeMutedWord(value: String): Outcome
```
tunjid: Its a bit awkward that `add` takes a `MutedWordPreference` and `remove` takes a s `String`. In any case, I think there should be a single `suspend updateMutedWords(update: TheUpdateClass): Outcome` here. This update should be sent after the user is done updating their muted words.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/MutedWordSheetState.kt
```diff
@@ -1,113 +1,1025 @@
...
+    data class UpdateDuration(val duration: Instant?) : Action("UpdateDuration")
+    data class UpdateTargets(val targets: List<String>) : Action("UpdateTargets")
+    data class UpdateExcludeNonFollowers(val exclude: Boolean) : Action("UpdateExcludeNonFollowers")
+    object Load : Action("Load")
 }
 
-typealias MutedWordStateHolder =
-    ActionStateMutator<MutedWordAction, StateFlow<MutedWordState>>
+@Stable
+class MutedWordsStateHolder(
```
tunjid: This class seems redundant given the existence of `MutedWordsSheetState`. `MutedWordsSheetState` already has a `CoroutineScope` passed to it, it can be a mutator itself.

```kotlin
@Stable
class MutedWordsSheetState private constructor(
    private val stateHolder: MutedWordsStateHolder,
    private val coroutineScope: CoroutineScope,
    scope: BottomSheetScope,
) : 
    BottomSheetState(scope),
    ActionStateMutator<Action, StateFlow<State>> by scope.actionStateFlowMutator(...)
```

tunjid:
There's a lot of moving parts in this PR which make it a bit hard to review. That said, some things are apparent. 

* Most moderation state will be in the `Preferences` class.
* Reading `Preferences` once will provide all moderation settings needed for bottom sheets with a few exceptions.

With that in mind, it is more prudent to have `UserDataRepository` expose `val preferences : Flow<Preferences>` and just collect these preferences at the `ViewModel` level in the screen's `State`, then pass the muted words list to `MutedWordsSheetState`. The extra complexity of passing repositories to bottom sheet states is not worth it

Then, `MutedWordsSheetState` should have logic for editing the initial list passed to it. Any optimizations should be done at the `ViewModel` level and can be put off till later.


Also there's a bit of scope creep. There's a bunch of moderation UI being added with no implementation, like UI for blocked users which are non functional and will block bug fix releases. This PR should not add the Muted words state to any screen yet, till the core implementation is working. There also a few hard coded strings in the PR.

joelmuraguri:
This review was super helpful, thanks for calling this out. It helped clarify the intended flow a lot.

I’ve refactored the muted words feature so all add/remove/edit operations are handled purely in-memory within the `MutedWordsSheetState`, keeping UI feedback immediate and consistent with the MVI approach.

The repository now exposes a single `updateMutedWords(ReplaceAll)` path, which is invoked only on Save, removing incremental network calls and duplicated preference update logic. Preferences continue to flow into the sheet as inputs, while persistence is handled centrally via the repository on commit.

Changes have been pushed.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/Preferences.kt
```diff
@@ -136,5 +136,5 @@ data class MutedWordPreference(
     @Serializable
     value class Target(
         val value: String,
-    )
+    ) : UrlEncodableModel
```
tunjid: Do we still need this? We don't pass it as a route argument.

feature/gallery/src/commonMain/kotlin/com/tunjid/heron/gallery/GalleryViewModel.kt
```diff
@@ -52,6 +53,7 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.emitAll
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.merge
```
tunjid: Seems like an unused import.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/MutedWordSheetState.kt
```diff
@@ -1,113 +1,916 @@
...
+}
+
+@Stable
+class MutedWordsSheetState(
+    private val coroutineScope: CoroutineScope,
+    scope: BottomSheetScope,
+    preferences: Flow<Preferences>,
+    private val onSave: suspend (List<MutedWordPreference>) -> Unit,
+) : BottomSheetState(scope),
+    ActionStateMutator<Action, StateFlow<State>> by coroutineScope.actionStateFlowMutator(
```
tunjid: Thanks for simplifying this. Since we intend to have a reference to the `Preferences` in the `State` of the calling screen, perhaps there is no need for a mutator either. What do you think about keeping the convention about just managing bottom sheet states with Compose state directly with no mutators?

We considered mutators bc we wanted to pass flows to the bottom sheet states for lazy loading, but we saw that was unnecessarily complicated and unwieldily. A decent compromise would be passing states directly and lazy load only when the sheet is showing.

We could have:

```kotlin
class MutedWordsSheetState(
    private val coroutineScope: CoroutineScope,
    scope: BottomSheetScope,
) {
    var mutedWords by mutableStateOf(emptyList<MutedWordPreference>())
}

companion object {
        @Composable fun rememberMutedWordsSheetState(
            mutedWords: List<MutedWordsPreference>,
            onShown: () -> Unit,
            onSave: (List<MutedWordPreference>) -> Unit,
        ): MutedWordsSheetState {
            val coroutineScope = rememberCoroutineScope()
            val state = rememberBottomSheetState { scope ->
                MutedWordsSheetState(
                    coroutineScope = coroutineScope,
                    scope = scope,
                )
            }.also { it.mutedWords = mutedWords}
            MutedWordsBottomSheet(state, onShown, onSave)
            return state
        }
}
```

Note:

* We should not hold a reference to the `onShown` or `onSave` callback in the state. It should be passed to the `MutedWordsBottomSheet` Composable. `rememberBottomSheetState` is a __capturing lambda__ and you will capture a stall callback passing it to the state.
* `onSave` should not suspend, it should be synchronous. The `WriteQueue` will suspend and deal with back off.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/Preferences.kt
joelmuraguri: No, its unnecessary

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/MutedWordSheetState.kt
tunjid: Feel free to keep using a Mutator if you prefer, though the capturing lambdas should be updated and we shouldn't pass a `Flow` or suspending method.

joelmuraguri: This makes sense. Given that muted word edits are now fully in-memory and committed only on Save, the mutator and Flow input are no longer pulling their weight. 
Working on the changes

PR #757, opened 2025-12-29
Muted word moderation UI implementation

joelmuraguri:


data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
```diff
@@ -140,3 +92,5 @@ internal class OfflineUserDataRepository @Inject constructor(
             }
         }.toOutcome()
 }
+
+fun UserDataRepository.readPreferences(): Flow<Preferences> = preferences
```
tunjid: Do we need this extension method? It just calls `preferences`.

feature/feed/src/commonMain/kotlin/com/tunjid/heron/feed/FeedViewModel.kt
```diff
@@ -141,6 +150,14 @@ fun recentConversationMutations(
             copy(recentConversations = conversations)
         }
 
+private fun loadPreferencesMutations(
+    userDataRepository: UserDataRepository,
+): Flow<Mutation<State>> =
+    userDataRepository.readPreferences()
+        .mapToMutation {
+            copy(preferences = it)
+        }
```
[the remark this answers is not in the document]
tunjid: This is more effort than it is worth especially with generics.

feature/feed/src/commonMain/kotlin/com/tunjid/heron/feed/FeedViewModel.kt
```diff
@@ -228,6 +231,19 @@ private fun Flow<Action.UpdateFeedGeneratorStatus>.feedGeneratorStatusMutations(
...
+): Flow<Mutation<State>> =
+    mapToManyMutations { action ->
+        writeQueue.enqueue(
+            Writable.TimelineUpdate(
+                Timeline.Update.OfMutedWord.ReplaceAll(
+                    mutedWordPreferences = action.mutedWordPreferences,
+                ),
+            ),
+        )
+    }
```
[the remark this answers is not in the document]
tunjid: This is a good candidate for the ui.timeline module, but we should consider it in a separate PR later. This is bc we may need to show error mutations and we'd want to make sure thats well designed.

feature/gallery/src/commonMain/kotlin/com/tunjid/heron/gallery/GalleryViewModel.kt
```diff
@@ -137,6 +147,14 @@ private fun loadPostMutations(
     )
 }
 
+private fun loadPreferencesMutations(
+    userDataRepository: UserDataRepository,
+): Flow<Mutation<State>> =
+    userDataRepository.readPreferences()
```
tunjid: should just be `preferences`.

feature/moderation/src/commonMain/kotlin/com/tunjid/heron/moderation/ModerationScreen.kt
```diff
@@ -235,6 +251,31 @@ private fun LazyListScope.subscribedLabelersSection(
     )
 }
 
+private fun LazyListScope.moderationToolsMenuSection(
+    onMutedWordsClicked: () -> Unit,
+) {
+    item(
+        key = Res.string.moderation_options_title.key,
+    ) {
+        SectionTitle(
```
tunjid, line 260: I think the `SectionTitle` should be inside its own `item` block. Is there a reason you wanted to but both the title and item in the same card?

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostOptionsSheetState.kt
```diff
@@ -149,18 +155,8 @@ sealed class PostOption {
...
-        data object MuteWords : Moderation(
-            title = Res.string.mute_words_tags,
-            icon = Icons.Default.FilterAlt,
-        )
-
-        data object BlockUser : Moderation(
-            title = Res.string.block_user,
-            icon = Icons.Rounded.Block,
-        )
+    sealed class Moderation : PostOption() {
```
tunjid, line 158: Why remove the values in these? It's likely for `BlockUser` you will need to pass the `ProfileId` of the profile to block. At that point it will become a data class

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
```diff
@@ -96,41 +83,6 @@ internal class OfflineUserDataRepository @Inject constructor(
         copy(useCompactNavigation = compactNavigation)
     }
 
-    override suspend fun updateMutedWords(
```
tunjid, line 99: Why delete this?

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
```diff
@@ -42,16 +35,10 @@ interface UserDataRepository {
     suspend fun setCompactNavigation(
         compactNavigation: Boolean,
     ): Outcome
-
-    suspend fun updateMutedWords(
```
tunjid, line 46: Why was this deleted?

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
tunjid: oh I see, its now unused. The updates go through timeline repository.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
tunjid: I see, it goes through timeline repository.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
joelmuraguri: Just followed a common pattern in the project of introducing helper ext. Makes sense to scrap it out if it an overkill

feature/gallery/src/commonMain/kotlin/com/tunjid/heron/gallery/GalleryViewModel.kt
joelmuraguri: Noted

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostOptionsSheetState.kt
joelmuraguri: After introducing `ModerationItemRow`, there was no need for the `icon` and `title ` values in the sealed class. Though I'll adjust it in later PR in `BlockerUser`

feature/moderation/src/commonMain/kotlin/com/tunjid/heron/moderation/ModerationScreen.kt
joelmuraguri: Yes this is right.

feature/feed/src/commonMain/kotlin/com/tunjid/heron/feed/FeedViewModel.kt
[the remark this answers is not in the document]
joelmuraguri: Seeing this a little to late after making commits.  Could we have the separate PR for this as a priority, as this pattern seems might be repeated in Moderation features.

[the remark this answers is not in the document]
tunjid: I'm not sure it should be a priority bc the current Gemini suggestion isn't helpful, its just noisy.

* The generic type `T` just reads a `List<MutedWordPreference>`, it doesn't perform any extra logic with the generic type.
* There is no real mutation being applied to the generic type `S`. Even if there was, it would look really complicated bc `S` is more often than not a data class and would end up calling a `copy` method that already exists. It would only make sense if `S` had a data class field that the method applied. For example `CursorQuery` is an interface that exposes `CusorQuery.Data`.

i.e unless the state `S` is is actually being mutated (like in `CursorQuery`) the generics just add noise since all its doing is enqueueing the write.

Here's an example of why this is difficult to do right. The following code block exists in a lot of view models bc each time, the State is being mutated with an error message. 

```kotlin
private fun Flow<Action.SendPostInteraction>.postInteractionMutations(
    writeQueue: WriteQueue,
): Flow<Mutation<State>> =
    mapToManyMutations { action ->
        when (writeQueue.enqueue(Writable.Interaction(action.interaction))) {
            WriteQueue.Status.Dropped -> emit {
                copy(messages = messages + action.interaction.failedWriteMessage())
            }
            WriteQueue.Status.Duplicate -> emit {
                copy(messages = messages + action.interaction.duplicateWriteMessage())
            }
            WriteQueue.Status.Enqueued -> Unit
        }
    }
```

A reusable generic form for any action and state pair would be:

```kotlin
fun <T, S> Flow<T>.postInteractionMutations(
    writeQueue: WriteQueue,
    interactionMapper: T.() -> Post.Interaction,
    stateMemoMutation: S.(Memo) -> S,
): Flow<Mutation<S>>  = mapToManyMutations { action ->
    val interaction = action.interactionMapper()
    when (writeQueue.enqueue(Writable.Interaction(interaction))) {
        WriteQueue.Status.Dropped -> emit {
            stateMemoMutation(interaction.failedWriteMessage())
        }
        WriteQueue.Status.Duplicate -> emit {
            stateMemoMutation(interaction.duplicateWriteMessage())
        }
        WriteQueue.Status.Enqueued -> Unit
    }
}
```

We don't have that here. This is the generic form I'd like for a future PR and is out of scope for this one. We'd like to enforce a pattern for any outcome of enqueueing things to write provide feedback to the user if necessary.

So let's leave the generics out of this PR for now, design the above to work for `Writable` types in general, then revisit.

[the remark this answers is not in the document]
joelmuraguri: Thanks for this clarification!

PR #777, opened 2025-12-31
feat: migrate from bottom nav to M3E toolbar

### LINKED ISSUE #773: Migrate from navbar + fab to M3E flexible toolbar
Using the new Material 3 Expressive Flexible Toolbar would look a lot cleaner than the current bottom navbar + FAB. 

Migrating from the bottom navbar/fab to a flexible toolbar

Compact bottom nav bar

**Additional context**
https://m3.material.io/components/toolbars/overview

ThatOneCalculator:
Closes #773 
Mostly done by Claude

https://github.com/user-attachments/assets/ab5fc39a-0516-4aec-ab3b-de0cb281fce4

### DIFF
--- a/scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/scaffold/PaneFloatingToolbar.kt
+++ b/scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/scaffold/PaneFloatingToolbar.kt
@@ -200,7 +200,7 @@ private fun NavItemButton(
                         tint = if (selected) {
                             MaterialTheme.colorScheme.onPrimary
                         } else {
-                            MaterialTheme.colorScheme.onSurface
+                            MaterialTheme.colorScheme.onSurfaceVariant
                         },
                     )
                 },


ThatOneCalculator:
~~Drafting as there's some weird color issues I'm currently fixing~~ fixed

tunjid:
I actually experimented with this during development and I don't think the M3E toolbar is the right navigation paradigm for heron bc it makes the bottom nav design behavior too coupled with the FAB. Heron also has first class supports tablets. The M3E toolbar doesn't have a good design story for how the fab behaves here.

This is what Heron looks like on tablets atm: 

https://github.com/user-attachments/assets/bee44e06-bdff-4ca7-90b6-bb2d3d7d794d

ThatOneCalculator:
I think a reasonable compromise would be toolbar on mobile and navbar + fab on tablet then?

tunjid:
I'm not sure its a compromise tbh. In the [M3 expressive video](https://www.youtube.com/watch?v=n17dnMChX14) the UI that is updated to the M3E toolbar was a regular toolbar which a bottom nav is not.

Screenshot 2026-01-02 at 9 54 59 PM

Screenshot 2026-01-02 at 9 55 33 PM

Design is subjective, but here are my other issues with it:

* The Fab is now the only element in the app with that hue and I find it a distracting. Its the primary action, it should have the apps primary color.
* It requires using `FloatingToolbarScrollBehavior`, the previous implementation had a pixel perfect offset on scroll which matched with the toolbar. The M3E version fires off a scroll anim. This doesn't match the design spec.
* The touch targets for the nav items are a bit to close.  My current roadmap may add a 5th item in the bottom nav. The M3E one isn't flexible enough for this.

Ultimately, I can't justify this change. Feel free to reopen if you have other opinions if you want to discuss them, I'm going to close this for now.

ThatOneCalculator:
I agree with some of the points, but the main one I will rebut is purpose – both Google Chat and Google Photos use the toolbar as the primary means of navigation between pages.

tunjid:
@ThatOneCalculator Google photos for me seems to still use a bottom nav bar. Material does have an annoyng way of changing names, the docs say its not called a flexible toolbar. Though the width in google photos spans the entire screen

PR #795, opened 2026-01-04
feat: Appearance settings (Dynamic Theming)

### LINKED ISSUE #770: Material You colors
The app is designed with Material 3 Expressive mostly, but the lack of Material You/dynamic colors is a bit disappointing.

The app to use the device's Material You colors

N/A

ThatOneCalculator:
Closes #770

I'm unsure if the original concerns about contrast/legibility apply anymore, this looks pretty fine to me.

| Light | Dark |
|-------|------|
| Screenshot_20260103-164618 | Screenshot_20260103-164623  |
| Screenshot_20260103-164649 | Screenshot_20260103-164645 |
| Screenshot_20260103-164658 | Screenshot_20260103-164701 |
| Screenshot_20260103-164718 | Screenshot_20260103-164722 |
| Screenshot_20260103-164733 | Screenshot_20260103-164736 |

### DIFF
--- a/composeApp/src/commonMain/kotlin/com/tunjid/heron/di/AppGraph.kt
+++ b/composeApp/src/commonMain/kotlin/com/tunjid/heron/di/AppGraph.kt
@@ -21,6 +21,7 @@ import com.tunjid.heron.conversation.di.ConversationBindings
 import com.tunjid.heron.data.di.AppCoroutineScope
 import com.tunjid.heron.data.di.DataBindings
 import com.tunjid.heron.data.repository.AuthRepository
+import com.tunjid.heron.data.repository.SavedStateDataSource
 import com.tunjid.heron.data.utilities.writequeue.WriteQueue
 import com.tunjid.heron.editprofile.di.EditProfileBindings
 import com.tunjid.heron.feed.di.FeedBindings
@@ -97,6 +98,7 @@ interface AppGraph {
         authRepository: AuthRepository,
         navigationStateHolder: NavigationStateHolder,
         notificationStateHolder: NotificationStateHolder,
+        savedStateDataSource: SavedStateDataSource,
         imageLoader: ImageLoader,
         videoPlayerController: VideoPlayerController,
         writeQueue: WriteQueue,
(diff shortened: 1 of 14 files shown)


tunjid:
IDK, some of the green ones seem hard to read to me. That said, this is subjective and I'm okay with this change as long as it is opt in. The change is significant enough to not have dynamic theming on by default, especially since it takes away from a brand identity.

To make it opt in:

* User settings are saved in the `SavedStateDataStore`: https://github.com/tunjid/heron/blob/main/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/SavedStateDataSource.kt#L186-L193
* In the preferences field, you can add (with `@ProtoNumber(9)`) a `useDynamicTheming` option, it will need to default to false as not all platforms support this option, and the default user experience should be the existing behavior.
* In the Settings screen, add a toggle for switching this on and off. You will have to make sure this setting only shows for Android. The current way things like this are done in the app is to use `expect/actual` to have different implementations for each platform.

Its a fair amount of stuff to add to support it, and its currently deprioritized as my focus is on moderation, and adding things like blocking, muting etc. You seem to use claude, so I imagine it should be trivial for it to knock it out.

ThatOneCalculator:
Sounds good, I'll probably stick it under a new "Appearance" category in settings.

ThatOneCalculator:
@tunjid 

https://github.com/user-attachments/assets/7f3ed8d0-79c9-4f68-bb7a-50b3b3854d2d

tunjid:
@ThatOneCalculator ty! I'll pull your changes into a new branch tomorrow and run them to get CI passing. I'm not sure why the build job keeps failing. I'll test this approach on your app name PR first.

tunjid:
@ThatOneCalculator when this goes out in the next release, do you mind if I add these screenshots to a post and tag you?

ThatOneCalculator:
> @ThatOneCalculator when this goes out in the next release, do you mind if I add these screenshots to a post and tag you?

I don't mind at all! Thank you :)

I might have one more improvement coming in a couple minutes in another PR that depends on this one too

tunjid:
Ty!

PR #797, opened 2026-01-04
feat: Auto-hide navbar setting

ThatOneCalculator:
Right now, the navbar always auto-hides when scrolling, but this gives me a bit of a headache (probably motion sensitivity?) when scrolling (and many other apps don't have this behavior), so I added an optional toggle to disable this behavior.

https://github.com/user-attachments/assets/6de0ca8e-b125-48a4-9d9c-5eb9336d15b8

### DIFF
--- a/feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/di/Bindings.kt
+++ b/feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/di/Bindings.kt
@@ -228,8 +228,8 @@ class ProfileBindings(
                                         isSignedOut -> signInDestination()
                                         else -> composePostDestination(
                                             type =
-                                                if (state.isSignedInProfile) Post.Create.Timeline
-                                                else Post.Create.Mention(state.profile),
+                                            if (state.isSignedInProfile) Post.Create.Timeline
+                                            else Post.Create.Mention(state.profile),
                                             sharedElementPrefix = null,
                                         )
                                     },


ThatOneCalculator:
@tunjid I think I figured out why this part is failing CI!

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':composeApp:processReleaseGoogleServices'.
> Malformed root json at /home/runner/work/heron/heron/composeApp/src/release/google-services.json
```

Forks don't have access to Github Secrets from the upstream repository, so the action step that populates `composeApp/src/release/google-services.json` puts in empty/invalid data.

tunjid:
You will probs have to rebase on UI changes: https://github.com/tunjid/heron/pull/798

I needed to drill the bottom nav bar height preference props to keep the math for hiding the bottom nav bar on scroll correct, and I had to pay the tech debt of exposing the `SavedStateDataSource`. Its supposed to be a dater layer impl detail.

ThatOneCalculator:
Gotcha, will take a look tmrw/Monday :)

ThatOneCalculator:
Also one issue I noticed with this PR is that scrolling should still hide the fab's label after scrolling (should the label even be shown at all...?)

scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/scaffold/AppState.kt
```diff
@@ -353,6 +353,6 @@ internal val LocalSplitPaneState = staticCompositionLocalOf<SplitPaneState> {
     throw IllegalStateException("No SplitPaneState provided")
 }
 
-internal val LocalAppState = staticCompositionLocalOf<AppState> {
```
tunjid, line 356: Oh this is an implementation detail I never want exposed to the wider app. Too big a can of worms.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/di/Bindings.kt
```diff
@@ -167,9 +169,14 @@ class ProfileBindings(
             val state by viewModel.state.collectAsStateWithLifecycle()
             val paneScaffoldState = rememberPaneScaffoldState()
 
+            val appState = LocalAppState.current
+
+            val autoHideNavigationBar = appState.preferences?.autoHideNavigationBar ?: true
```
tunjid: This leaks app state preferences logic, it needs to be abstracted away reasonably.

ThatOneCalculator:
Gotcha. Will implement the fixes by hand tomorrow, the majority of this PR was Claude'd

ThatOneCalculator:
@tunjid redid it in a much better way, akin to `isCompact` for `bottomNavigationNestedScrollConnection` :)

tunjid:
Did you try conditionally applying the nested scroll connection?

```kotlin
            paneScaffoldState.PaneScaffold(
                modifier = Modifier
                    .fillMaxSize()
                    .predictiveBackPlacement(paneScope = this)
                    .then(if(prefersBottomNavScroll) Modifier.nestedScroll(bottomNavigationNestedScrollConnection) else Modifier)
              )      
```

I don't think any of the changes in `AccumulatedOffsetNestedScrollConnectionExt.kt` need to be made. Its a conscious decision to couple the fab and bottom nav offsets tightly.

ThatOneCalculator:
> Its a conscious decision to couple the fab and bottom nav offsets tightly.

I get that, but the point is that the FAB should still shrink on scroll without being offset. If we do

```kt
                    .then(if(prefersBottomNavScroll) Modifier.nestedScroll(bottomNavigationNestedScrollConnection) else Modifier)
```

then the FAB won't shrink in-place while scrolling if `prefersBottomNavScroll` is false.

tunjid:
@ThatOneCalculator in that case, can it be derived from the `topAppBarNestedScrollConnection`? These are my concerns:

* The fab offset should be tightly coupled with the bottom nav's.
* `fabScrollConnection` is another class allocated with its snapshot state read on each frame.

The less objects allocated and snapshot states to read, maintain and keep in sync the better. Ditto the amount of modifiers applied. If adding this preference can improve app perf, I want to aggressively optimize for it.

Your preference does less, it should use less resources. `isExpanded` looks at the total scroll amount, this can be read from the top or bottom app bar scroll. `Offset` in compose has operator overloading. To use the `topAppBarNestedScrollConnection`, its `Offset` can be multiplied by `-1` and passed to `isFabExpanded`. This way, when your preference is applied:

* Less modifiers are in the tree.
* Only one `AccumulatedOffsetNestedScrollConnection` has its snapshot state read.

ThatOneCalculator:
I'm going to be honest here, my lack of skill with the stack here is definitely catching up to me -- I'd be fine promoting this to Claude if you want, but I wouldn't be able to properly review the code outside of verifying that it compiles & works on my phone. Given you have a solid idea of how it should be implemented, would it be okay to ask you to implement that refactor?

tunjid:
Yeah I can do that, just cleaning up some stuff with how API calls to bluesky is done when reading preferences. Their caching layer doesn't immediately reflect writes and it can't be relied on as a source of truth immediately after writing stuff to it.

ThatOneCalculator:
Tysm!!

tunjid:
Superceded by https://github.com/tunjid/heron/pull/797

PR #802, opened 2026-01-04
fix: smaller strings for dynamic theme setting

ThatOneCalculator:
Fixes this:

image

### DIFF
--- a/feature/settings/src/commonMain/composeResources/values/strings.xml
+++ b/feature/settings/src/commonMain/composeResources/values/strings.xml
@@ -31,6 +31,6 @@
     <string name="refresh_timelines_on_launch">Refresh timelines on launch</string>
     <string name="moderation">Moderation</string>
     <string name="appearance">Appearance</string>
-    <string name="use_dynamic_theming">Use app theme from wallpaper color</string>
+    <string name="use_dynamic_theming">Use device theme from wallpaper color</string>
     <string name="use_compact_navigation">Use compact bottom navigation bar</string>
 </resources>


ThatOneCalculator:
@tunjid even with the latest commit it looks a bit weird pushed over :/

tunjid:
Yeah, I verified on the ci build. I use ci to run a smoke test release build.

ThatOneCalculator:
@tunjid how about "App theme from wallpaper color"?

tunjid:
@ThatOneCalculator id like there to be a verb associated, I'd like to keep use. Text being multi-line is okay, just the layout and alignment needs fixing

ThatOneCalculator:
Added back "Use", still is more consise now :)

feature/settings/src/commonMain/composeResources/values/strings.xml
```diff
@@ -31,6 +31,6 @@
     <string name="refresh_timelines_on_launch">Refresh timelines on launch</string>
     <string name="moderation">Moderation</string>
     <string name="appearance">Appearance</string>
-    <string name="use_dynamic_theming">Use wallpaper influenced dynamic app theme</string>
+    <string name="use_dynamic_theming">Use app theme from wallpaper color</string>
```
tunjid: Should this be "device theme from wallpaper color"?

ThatOneCalculator: Either that, or just "Use theme from wallpaper color". I think both options ("device theme" or just "theme") work good, just lmk :)

ThatOneCalculator: Either that, or just "Use theme from wallpaper color". I think both options ("device theme" or just "theme") work good, just lmk :)

tunjid: "device theme" works for me, it shows it matches the rest of the device. The app has no themes of its own yet; I plan to add some later modeled after different kinds of herons.

PR #805, opened 2026-01-04
make FABs rectangular

ThatOneCalculator:
Aligns with MD3/Material Expressive :)

### DIFF
--- a/feature/compose/src/commonMain/kotlin/com/tunjid/heron/compose/ui/ComposePostFab.kt
+++ b/feature/compose/src/commonMain/kotlin/com/tunjid/heron/compose/ui/ComposePostFab.kt
@@ -18,10 +18,10 @@ package com.tunjid.heron.compose.ui
 
 import androidx.compose.animation.ExperimentalSharedTransitionApi
 import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.shape.CircleShape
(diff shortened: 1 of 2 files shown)


tunjid:
Heron has its own design system and while it derives from material expressive its distinct. Its initial inspiration is the crane material 2 design study: https://m2.material.io/design/material-studies/crane.html

Shapes are more organic and round, this is too much of a departure. Only cards in Heron have completely flat edges.

i.e, I have no desire to align with Material 3 Expressive fully, otherwise Heron becomes a cookie cutter M3E instead of an expression of its own design sensibility.

ThatOneCalculator:
While my personal preference leans towards the apps on my system to look more cohesive/uniform, I totally get where you're coming from. Would it be fine to make this an optional toggle in appearance settings for those who prefer the look of M3E?

tunjid:
Having that would mean Heron has no true design direction, and it can be entirely up to the user. Heron's tenet is that it's graceful, not that its super customizable. Another user may want a different shape for the fab, and that would have to be taken as a valid consideration too.

Opinionated design protects me as the maintainer from having to consider everything any number of users want. It also frees me to explore other designs I want to which is necessary for me as a passion project.

ThatOneCalculator:
Gotcha, I totally get where you're coming from on that.

PR #776, opened 2025-12-31
fix: match interaction colors

ThatOneCalculator:
- Matches interaction colors with notification interaction colors
- Animates interaction colors with a short 100ms fade instead of jumping

PR #809, opened 2026-01-04
post interaction button changes

ThatOneCalculator:
- Re-apply post interaction button change from #776 (got overridden with force push)
- Change `ArrowCircleUp` to `MoreVert` (3 dots) for `MoreOptions` button (re-up of #794)
- Add short pop animation & haptic to interact buttons

| Before | After |
|--------|--------|
| image | image |

Btw this will probably be my last PR/commit for today, will wrap up loose ends with the other PRs tomorrow :)

### DIFF
--- a/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
+++ b/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
@@ -16,13 +16,18 @@
 
 package com.tunjid.heron.timeline.ui.post
 
+import androidx.compose.animation.Crossfade
 import androidx.compose.animation.animateColorAsState
+import androidx.compose.animation.core.Spring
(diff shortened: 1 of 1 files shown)


tunjid:
This PR changes quite a bit, some we've already discussed and some new. Its should be split so we can discuss the merits of each individually.

* Share button icon: I'd like to hold off changes to this until moderation features are completed. @joelmuraguri is working on this.
* Haptics: This should be a user configurable option. The Preferences data structure might need some refactoring to accommodate this, and that option will need to be piped through the app layers to this button.
* Interaction button pop: I'm good with this change for now. The long term plan for this is to use lottie animations for this: https://lottiefiles.com/free-animations/like. However integrating lottie is on the backlog, there are higher priority things to work on. The pop animation is a good stop gap. The `scale` should be applied using `Modifier.graphicsLayer { }` for performance reasons however.

ThatOneCalculator:
To address each bullet point:
- Holding off on that
- Holding off on this as well until discussion in #810 is resolved
- Re-upped in #811

PR #811, opened 2026-01-05
post interaction button changes (re-up)

ThatOneCalculator:
Re-up of #809 just for the button size & pop

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
+++ b/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/post/PostInteractions.kt
@@ -69,8 +69,6 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.graphics.vector.rememberVectorPainter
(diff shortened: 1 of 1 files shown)


tunjid:
There are some small perf things I'll fix on top of your commits.

ThatOneCalculator:
Ty! I'm sure that I'm missing a lot of best practices, I've never done any meaningful work in Jetpack Compose until now.

tunjid:
Sorry still reviewing this. One of the reasons animating this has been put off for so long is that the architecture of the app is offline first and the database is always the source of truth. So changes to `isChecked` always has a bit of a delay while the round trip to network and back are pending.

When the app is offline, the anim is really delayed and looks off. The Bluesky post about assuming opportunistic writes would be nice, but then I have to manage state about what the expected value and the actual value are. So each tap has an expected value where the db should eventually catch up to. In the case the app is offline, this won't happen anytime soon and I need to have a sort of intermediate state showing the like (or bookmark, repost, etc) is pending. This is why I wanted to use lottie so I could play an animation loop till the action is confirmed.

ThatOneCalculator:
I don't think that the pop animation looping would look good, imo it should either be:

- Color change plays immediately, loading animation plays if >500ms to like, then pop animation when it goes through
- Color changes plays & haptic immediately on button push, then pop animation when it goes through (I prefer this one)

tunjid:
Superceded by the PR above. Closing.

PR #843, opened 2026-01-12
Block and Unblock account explanation in Profile screen implementation

joelmuraguri:


tunjid:
Let's implement this as a [pop up](https://developer.android.com/develop/ui/compose/components/menu) to the right of the `Follow` button. Use the same vertical 3 dot icon, but it should be the same height as the `Follow` button. Maybe use the same `FilterChip`.

It should not be a bottom sheet bc the user has to reach to the top to press it, and then reach down to act. Also since it is only used by the `ProfileScreen`, there is no need to define it all the way in the `:ui:timeline` module.

feature/profile/src/commonMain/composeResources/values/strings.xml
```diff
@@ -40,4 +41,5 @@
     <string name="label_off">Off</string>
     <string name="label_show_badge">Show badge</string>
     <string name="label_hide">Hide</string>
+    <string name="more_options">More Options</string>
```
tunjid: This should be defined in `ui:core` string resources. There are a lot of common string there. To use them across modules, use the `CommonStrings.more_options` import.

feature/profiles/src/commonMain/composeResources/values/strings.xml
```diff
@@ -17,4 +17,12 @@
 <resources>
     <string name="heron">Heron</string>
     <string name="back">Back</string>
-</resources>
\ No newline at end of file
+    <string name="likes">Likes</string>
+    <string name="reposts">Reposts</string>
```
tunjid: These string have already been defined in `CommonStrings`.

feature/profiles/src/commonMain/kotlin/com/tunjid/heron/profiles/di/Bindings.kt
```diff
@@ -279,6 +293,11 @@ class ProfilesBindings(
                 topBar = {
                     PoppableDestinationTopAppBar(
                         onBackPressed = { viewModel.accept(Action.Navigate.Pop) },
+                        title = {
+                            Text(
```
tunjid: This should be the `AppBarTitle` composable.

feature/profiles/src/commonMain/kotlin/com/tunjid/heron/profiles/di/Bindings.kt
```diff
@@ -298,3 +317,12 @@ class ProfilesBindings(
         },
     )
 }
+
+fun Load.titleRes(): StringResource = when (this) {
```
tunjid, line 321: These string resources should be pulled from `CommonStrings`.

ui/timeline/src/commonMain/composeResources/values/strings.xml
```diff
@@ -53,7 +53,14 @@
     <string name="expand_options">Expand options</string>
     <string name="remove_feed">Remove</string>
     <string name="mute_video">Mute video</string>
-    <string name="block_user">Block User</string>
+    <string name="block_user">Blocked Accounts</string>
+    <string name="block_account">Block Account</string>
```
tunjid: Since these strings are going to be used for both `ProfileScreen` and `PostActions` maybe they should be in common strings.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/ProfileActionsSheetState.kt
```diff
@@ -0,0 +1,70 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import com.tunjid.heron.timeline.utilities.BlockAccount
+import com.tunjid.heron.ui.sheets.BottomSheetScope
+import com.tunjid.heron.ui.sheets.BottomSheetScope.Companion.ModalBottomSheet
+import com.tunjid.heron.ui.sheets.BottomSheetScope.Companion.rememberBottomSheetState
+import com.tunjid.heron.ui.sheets.BottomSheetState
+
+@Stable
+class ProfileActionsSheetState(
```
tunjid: We should be able to delete this class.

joelmuraguri:
> Let's implement this as a [pop up](https://developer.android.com/develop/ui/compose/components/menu) to the right of the `Follow` button. Use the same vertical 3 dot icon, but it should be the same height as the `Follow` button. Maybe use the same `FilterChip`.
> 
> It should not be a bottom sheet bc the user has to reach to the top to press it, and then reach down to act. Also since it is only used by the `ProfileScreen`, there is no need to define it all the way in the `:ui:timeline` module.

Makes sense 👍 
A popup menu anchored to the right side of Follow button will give better reachability and fits the single-action use case. I’ll switch this from a bottom sheet to a popup menu using the same 3-dot icon (matched height with Follow / FilterChip) and scope it directly to `ProfileScreen` instead of  `:ui:timeline` module.

feature/profiles/src/commonMain/composeResources/values/strings.xml
joelmuraguri: I've not seen these strings in `ui:core` as already defined. 
I`ll add them

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ui/ProfileActionsMenu.kt
```diff
@@ -0,0 +1,129 @@
...
+    )
+}
+
+@Stable
+internal data class ProfileActionMenuItem(
+    val title: String,
+    val icon: ImageVector,
+    val isDestructive: Boolean = false,
+    val showDividerBelow: Boolean = false,
+    val onClick: () -> Unit,
```
tunjid, line 128: lambdas shouldn't be in data classes, the generated equals and hashcode will not be stable.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/State.kt
```diff
@@ -195,6 +196,17 @@ sealed class Action(val key: String) {
         val target: LinkTarget,
     ) : Action(key = "BioLinkClicked")
 
+    data class BlockAccount(
```
tunjid, line 199: These two mutations should have the same key, otherwise they will run cuoncurrently

tunjid (commented):
Merging as is to unblock myself, but take note of the suggested changes.

PR #847, opened 2026-01-14
Add mutes to profile restrictions

tunjid:
cc @joelmuraguri just made a few changes and added mute support.


joelmuraguri:
> cc @joelmuraguri just made a few changes and added mute support.

Thanks for the help. 
This PR is really insightful with how ProfileActionMenu, Moderation actions have been refactored.

PR #848, opened 2026-01-14
Implement Profile restrictions across timeline

joelmuraguri:


feature/home/src/commonMain/kotlin/com/tunjid/heron/home/HomeScreen.kt
```diff
@@ -338,14 +341,18 @@ private fun HomeTimeline(
         },
         onShown = {},
     )
+    var pendingModeration by remember {
```
tunjid: This is a good start but we can reuse so much of this code. There is little difference between a bottom sheet and a dialog. This can be:

```
@Composable
fun rememberModerationDialogState(
    action: PostAction.Moderation,
    onApproved: (PostAction.Moderation) -> Unit,
): ModerationDialogState {
    val state = remember { ModerationDialogState(...) }

    if(state.show) ModerationDialog(
         onApproved = onApproved,
    )

    return state
}
```

Then the `ModerationDialogState` is very similar to the bottom sheet states we have. Except it holds a single boolean, and if it is true.

Would keep the pattern we have and prevent so much copy pasting.

tunjid: We should be able to use the same state for blocks or mutes, the `PostAction` has enough signal to let us know what strings to use.

joelmuraguri: Thanks for the keen eye on this refactor, agreed on keeping dialog state aligned with our bottom sheet patterns to avoid duplication.
In recent push, `ProfileRestrictionsDialogState` and `ModerationDialogState` have been included.

tunjid:
@joelmuraguri your changes were good. I made a few changes on top:

* There was no need to add `PostAction.Moderation`, we already had `PostOption.Moderation`. I realize in my suggestion snippet above I wrote `PostAction.Moderation`, however `PostOption.Moderation` was the better destination. Feel free to have a conversation with me if I suggest something, it doesn't have to be verbatim. The suggestion was by looking at the PR, I didn't have IDE context.
* Your name for `ProfileRestrictionDialogState` in the profile screen is better than `ModerationDialogState`. I've used that too.

joelmuraguri:
> @joelmuraguri your changes were good. I made a few changes on top:
> 
> * There was no need to add `PostAction.Moderation`, we already had `PostOption.Moderation`. I realize in my suggestion snippet above I wrote `PostAction.Moderation`, however `PostOption.Moderation` was the better destination. Feel free to have a conversation with me if I suggest something, it doesn't have to be verbatim. The suggestion was by looking at the PR, I didn't have IDE context.
> * Your name for `ProfileRestrictionDialogState` in the profile screen is better than `ModerationDialogState`. I've used that too.

Well noted, I’ll make sure to engage more.
Appreciate the feedback!

PR #871, opened 2026-01-17
Data layer Create Notification channels impl

joelmuraguri:


data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/Preferences.kt
```diff
@@ -53,6 +53,8 @@ data class Preferences(
     val postInteractionSettings: PostInteractionSettingsPreference? = null,
     @ProtoNumber(11)
     val verificationPreferences: VerificationPreference? = null,
+    @ProtoNumber(12)
+    val notificationPreferences: NotificationPreferences? = null,
```
tunjid: Let's not keep notification preferences here. Instead, lets put them in `SavedState.Notifications`. The existing `Preferences` are mapped from a different network response.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/preferenceupdater/PreferenceUpdater.kt
```diff
@@ -71,6 +73,11 @@ internal interface PreferenceUpdater {
         networkPreferences: List<PreferencesUnion>,
         update: Timeline.Update,
     ): List<PreferencesUnion>
+
+    suspend fun update(
```
tunjid: Its probably best we don't mix these two preference types.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/SavedStateDataSource.kt
```diff
@@ -173,6 +174,7 @@ abstract class SavedState {
         val lastRead: Instant? = null,
         val lastRefreshed: Instant? = null,
         val hasPreviouslyRequestedPermissions: Boolean = false,
+        val preferences: NotificationPreferences? = null,
```
tunjid, line 177: This is fine, we should store it here. No need to also store it in the other preferences.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/NotificationPreferences.kt
```diff
@@ -0,0 +1,121 @@
...
+
+            fun allowsAuthor(isFollowed: Boolean): Boolean =
+                when (include) {
+                    Include.All -> true
+                    Include.Follows -> isFollowed
+                    Include.Unknown -> false
+                }
+
+            @Serializable
+            enum class Include(val value: String) {
```
tunjid: Let's make this a value class in case more options are added by the backend in the future. We don't want to fail on deserialization.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/NotificationPreferences.kt
```diff
@@ -0,0 +1,121 @@
...
+            val preference: Preference.Filterable,
+        ) : Update()
+
+        @Serializable
+        data class Simple(
+            override val reason: Reason,
+            val preference: Preference.Simple,
+        ) : Update()
+
+        @Serializable
```
tunjid: We own the `NotificationReason` enum so we can reuse it, there's no need to define a new one.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/NotificationPreferences.kt
```diff
@@ -0,0 +1,121 @@
...
+                Follows("follows"),
+                Unknown("unknown"),
+            }
+        }
+    }
+
+    @Serializable
+    sealed class Update {
+        abstract val reason: Reason
+
```
tunjid: To simplify things, the update class should be a simple data class. If more preference options are added in the future, we don't want to do a huge udate migration.

It can just be:

```kotlin

    @Serializable
    data class Update(
         val reason: Reason,
         val list: Boolean,
         val push: Boolean,
         val include: Include?,
     )
```

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/NotificationPreferences.kt
```diff
@@ -0,0 +1,121 @@
...
+            object Quote : Reason("quote")
+            object Reply : Reason("reply")
+            object Repost : Reason("repost")
+            object RepostViaRepost : Reason("repostViaRepost")
+            object StarterpackJoined : Reason("starterpackJoined")
+            object SubscribedPost : Reason("subscribedPost")
+            object Unverified : Reason("unverified")
+            object Verified : Reason("verified")
+        }
+    }
```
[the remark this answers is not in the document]
tunjid: This suggestion can be ignored, we don't control the backend so this is too rigid.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/writequeue/Writable.kt
```diff
@@ -161,6 +162,21 @@ sealed interface Writable {
         override suspend fun WriteQueue.write(): Outcome =
             profileRepository.updateProfile(update)
     }
+
+    @Serializable
+    data class NotificationUpdate(
+        val update: NotificationPreferences.Update,
+    ) : Writable {
+
+        override val queueId: String
```
tunjid, line 171: With the `Update` being a data class, this can now be `"notification-pref-$update"`

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -295,6 +306,77 @@ internal class OfflineNotificationsRepository @Inject constructor(
         ).toOutcome()
     } ?: expiredSessionOutcome()
 
+    override suspend fun updateNotificationPreferences(
+        update: NotificationPreferences.Update,
+    ): Outcome =
+        networkService.runCatchingWithMonitoredNetworkRetry {
+            getPreferencesForNotification()
+        }.toOutcome { it ->
```
tunjid: Using a simple update class should simplify this. The preference should just read the fields it needs from the `Update`.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/NotificationPreferences.kt
joelmuraguri: Thanks for pointing that out.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/NotificationPreferences.kt
joelmuraguri: Makes sense. 
This fix will cut down on repetition

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/NotificationPreferences.kt
joelmuraguri: I agree!
This will simplify the `updateNotificationPreferences` method

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -295,6 +306,78 @@ internal class OfflineNotificationsRepository @Inject constructor(
...
+                    unverified = if (update.reason == Notification.Reason.Unverified) simplePref else currentPrefs.unverified,
+                    verified = if (update.reason == Notification.Reason.Verified) simplePref else currentPrefs.verified,
+                    chat = currentPrefs.chat,
+                )
+            }
+
+            networkService.runCatchingWithMonitoredNetworkRetry {
+                putPreferencesV2(request = updateRequest)
+            }.toOutcome {
+                val notifications = savedStateDataSource
```
tunjid, line 369: This method is not aware of the signed in user and may continue running if the active account is changed. The top of the block should use `savedStateDataSource.inCurrentProfileSession { }`.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -295,6 +306,78 @@ internal class OfflineNotificationsRepository @Inject constructor(
...
+                    .savedState.value
+                    .signedInProfileData
+                    ?.notifications
+                    ?: SavedState.Notifications()
+
+                val updatedNotificationPreferences = preferenceUpdater.update(
```
tunjid: I don't think this method should call to preferences updater, notification preferences are separate from general preferences. Let's leave `PreferenceUpdater` untouched in this PR. `PreferenceUpdate` works with preferences from `app.bsky.preferences`, notification preferences work with `ap.bsky.notification.preferences`.

joelmuraguri: I think we can have
`NotificationPreferenceUpdater` exclusively for `app.bsky.notification.preference`. 

This approach offers separation of concerns.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
joelmuraguri: This is fixed in recent push

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -295,6 +307,79 @@ internal class OfflineNotificationsRepository @Inject constructor(
         ).toOutcome()
     } ?: expiredSessionOutcome()
 
+    override suspend fun updateNotificationPreferences(
+        update: NotificationPreferences.Update,
+    ): Outcome = savedStateDataSource.inCurrentProfileSession { signedInProfileId ->
+        if (signedInProfileId == null) return@inCurrentProfileSession expiredSessionOutcome()
+        networkService.runCatchingWithMonitoredNetworkRetry {
+            getPreferencesForNotification()
+        }.toOutcome { it ->
```
tunjid: `toOutcome` is typically used as the last step after a `Result` has been obtained. Since you're mapping, I think `mapToResult` to change one `Result` type to another is better.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -295,6 +307,79 @@ internal class OfflineNotificationsRepository @Inject constructor(
...
+            val currentPrefs = it.preferences
+            val updateRequest = if (update.include != null) {
```
tunjid: Having performed a null check, using a nullable `let` feels awkward. Use `include` as the subject of the `when` to avoid the compiler telling you it can't guarantee nullability.

```kotlin
val updateRequest =  when(val include = update.include) {
    null -> {
        val simplePref = Preference(
               list = update.list,
               push = update.push,
           )
        ...
    }
   else -> {
        val filterablePreference = FilterablePreference(
            include = FilterablePreferenceInclude.safeValueOf(includeValue),
            list = update.list,
            push = update.push,
        )
        ...
    }
}
```

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -467,6 +552,35 @@ internal class OfflineNotificationsRepository @Inject constructor(
                 }
         }
 
+    suspend fun loadNotificationPreferences() {
```
tunjid: This should probably be private, or deleted if unused.

data/database/schemas/com.tunjid.heron.data.database.AppDatabase/33.json
```diff
@@ -0,0 +1,3215 @@
+{
```
tunjid, line 1: You may have to rebase, this file was checked in this week.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
joelmuraguri: Got it though this method is intended to load `app.sky.notification.preference`

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
joelmuraguri: I agree.
Thanks for that clarification

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
joelmuraguri: You're right, the method should be removed. Instead `saveNotificationPrefrences` is in `AuthRepository`, this ensure we have access to `NotificationPreferences` at any point, for a signed in Profile

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
joelmuraguri: Thanks for this, that `Include` nullable is handled well for `FilterablePreference`  which are non nulls

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/utilities/WriteQueueStatusMemos.kt
```diff
@@ -55,7 +55,7 @@ fun Writable.writeStatusMessage(
     else -> {
         val isDropped = status == WriteQueue.Status.Dropped
         when (this) {
-            is Writable.Connection -> Memo.Resource(
+            is Writable.Connection -> Resource(
```
tunjid: Please keep the Memo prefix in these

PR #889, opened 2026-01-22
Build notification settings screen

joelmuraguri:


scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/navigation/NavigationMutator.kt
```diff
@@ -239,6 +239,11 @@ fun settingsDestination(): NavigationAction.Destination = pathDestination(
     referringRouteOption = ReferringRouteOption.Current,
 )
 
+fun notificationSettingsDestination(): NavigationAction.Destination = pathDestination(
+    path = "/notification-settings",
```
tunjid: This url should be `settings/notifications` to match https://bsky.app/settings/notifications

feature/notifications/src/commonMain/kotlin/com/tunjid/heron/notifications/di/Bindings.kt
```diff
@@ -161,13 +165,13 @@ class NotificationsBindings(
                             )
                         },
                         actions = {
-                            androidx.compose.animation.AnimatedVisibility(
```
tunjid, line 164: Shouldn't this be a row where both icons are side by side? You seem to have deleted the button for granting notification permisssions.

joelmuraguri: Good catch, the permission request action shouldn’t replace the settings action. Now we have them grouped in a row so settings action is always visible and the request button appears alongside it when permissions are missing.

Issue #780, opened 2026-01-01
Create notification channels when notification permission is granted


ianhanniballake:
**Is your feature request related to a problem? Please describe.**
As a follow up to #395, I noticed that after granting the notification permission, there was no way to adjust what notifications I want to receive until I actually get a notification of each type.

**Describe the solution you'd like**
It would be nice if every notification channel would be created as soon as the permission was granted, thus allowing users to immediately adjust what notifications they want to receive.

Ideally, the UI would also use an Intent with the [`Settings.ACTION_APP_NOTIFICATION_SETTINGS`](https://developer.android.com/reference/android/provider/Settings#ACTION_APP_NOTIFICATION_SETTINGS) action and [`Settings.EXTRA_APP_PACKAGE`](https://developer.android.com/reference/android/provider/Settings#EXTRA_APP_PACKAGE) to deep link from the UI directly into the notification settings page.

**Describe alternatives you've considered**
A more full featured in-app UI would perhaps be more approachable to users.

tunjid:
@joelmuraguri work on this next. You will need to:

* Update `SavedState.ProfileData.Notifications` and add a preferences field. 
    * It should be nullable and have a default value of null. This allows for seamless migration.
    *  It should be a sealed class. There are two types. Look at the API response for `getPreferencesForNotification()` to see the two types. Ignore the chat preference for now as Heron cannot and does not send notifications for chats.
* Update `NotificationsRepository` to fetch and persist notification settings. 
* Build the notifications settings screen.
    * Unlike the bluesky app, taping a preference should expand to show the options instead of navigating to another screen. Use the same expand/collapse logic as in the settings screen. Though you don't have to reuse the existing code.
* Allow users update notification settings.
* Check notification settings before pinging a user when a notification is received. When a notification is received it gets a payload. See this [code block](https://github.com/tunjid/heron/blob/main/scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/notifications/NotificationStateHolder.kt#L71-L80). The payload also contains a `reason` key. You will need to pull this out of the payload to check wether or not the notification should be processed. The value of the key maps to the `Notification.Reason` enum, which should also map to the preferences key from the API.

Split this work up into different PRs.

- [x] The data layer PR
- [x] The UI layer PR
- [ ] The logic for checking if a user wants to see a notification.

[...]

PR #894, opened 2026-01-26
Notification ping check when notification is received

joelmuraguri:
Fixes https://github.com/tunjid/heron/issues/780

### DIFF
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
@@ -167,7 +167,6 @@ internal class OfflineNotificationsRepository @Inject constructor(
     private val networkMonitor: NetworkMonitor,
     private val savedStateDataSource: SavedStateDataSource,
     private val notificationPreferenceUpdater: NotificationPreferenceUpdater,
-    private val userDataRepository: UserDataRepository,
     httpClient: HttpClient,
 ) : NotificationsRepository {
 
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/SavedStateDataSource.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/SavedStateDataSource.kt
@@ -245,7 +245,7 @@ internal fun SavedState.signedProfilePreferencesOrDefault(): Preferences =
             null -> Server.BlueSky.endpoint
         }.let(::preferencesForUrl)
 
-fun SavedState.signedNotificationPreferencesOrDefault(): NotificationPreferences =
+internal fun SavedState.signedNotificationPreferencesOrDefault(): NotificationPreferences =
     signedInProfileData?.notifications?.preferences ?: NotificationPreferences.Default
 
 private fun SavedState.AuthTokens?.ifSignedIn(): SavedState.AuthTokens.Authenticated? =


scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/notifications/NotificationStateHolder.kt
```diff
@@ -176,14 +188,23 @@ private fun Flow<NotificationAction.RegisterToken>.registerTokenMutations(
...
+            val reason = action.reason ?: run {
+                logcat(LogPriority.WARN) {
+                    "Notification dropped, unknown reason from payload: ${action.payload[NotificationAtProtoReason]}"
+                }
+                return@flatMapMerge emptyFlow()
+            }
 
             flow {
+                val notificationPreferences = userDataRepository.notificationPreferences.first()
+
```
tunjid: The logic is good, but let's keep it contained in the `NotificationsRepository`. Let's update `NotificationsQuery.Push` to have the reason argument. Then in the `NotificationRepository`, which already has access to saved state and user preferences, perform the filtering. i.e it should not be `result.isSuccess` if the user is not interested in the notification.

We shouldn't need to add `UserDataRepository` here.

joelmuraguri: Thanks for the suggestion! We now have the `NotitificationPreference` filtering implementation in repository layer.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -394,138 +399,178 @@ internal class OfflineNotificationsRepository @Inject constructor(
...
+                                        reasonSubject = null,
+                                        isRead = false,
+                                        indexedAt = now,
+                                        associatedPost = resolvedRecord.post,
+                                        viewerState = viewerState,
+                                    )
+
+                                else -> throw UnknownNotificationException(query.recordUri)
+                            }
+                            else -> throw NotificationFilteredOutException(query.reason)
```
tunjid: This should be `true` and `else`. `null` for `resolvedRecord.post.viewerStats?.threadMuted` means a user has set no preference, so the notification should go through.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -394,138 +399,178 @@ internal class OfflineNotificationsRepository @Inject constructor(
...
+                        }
+
+                        is Repost -> when (resolvedRecord.post.viewerStats?.threadMuted) {
+                            true -> throw MutedThreadException(resolvedRecord.post.uri)
+                            false -> when (resolvedRecord.via) {
```
tunjid: Same here, the `else` is the right branching logic.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -394,138 +399,178 @@ internal class OfflineNotificationsRepository @Inject constructor(
...
+                                        viewerState = viewerState,
+                                    )
+
+                                else -> throw UnknownNotificationException(query.recordUri)
+                            }
+                            else -> throw NotificationFilteredOutException(query.reason)
+                        }
                     }
+
+                    val notificationPreferences =
```
tunjid: Repositories do not depend on repositories, it can cause circular dependency injection loops. This is why utilities like `RecordResolver`, `PreferenceUpdater` and others exist.

You should use `SavedStateDataSource.value.signedNotificationPreferencesOrDefault()` to read the preference directly.

joelmuraguri: Got it! I’ve updated the code to read notification preferences directly from `SavedStateDataSource.value.signedNotificationPreferencesOrDefault()`, removing any repository-to-repository dependency.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
joelmuraguri: Changes applied. The `else` branch is now correctly used to ensure proper handling when `threadMuted` is `null`

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
joelmuraguri: Thanks for the clarification! I’ve updated the branching logic so that `null` now correctly allows the notification to pass through, and the `else` branch handles the filtered case as suggested.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/SavedStateDataSource.kt
```diff
@@ -245,7 +245,7 @@ internal fun SavedState.signedProfilePreferencesOrDefault(): Preferences =
             null -> Server.BlueSky.endpoint
         }.let(::preferencesForUrl)
 
-internal fun SavedState.signedNotificationPreferencesOrDefault(): NotificationPreferences =
+fun SavedState.signedNotificationPreferencesOrDefault(): NotificationPreferences =
```
tunjid: This can still be internal; its internal to the data layer so repositories can access it. This should be the last change and this PR is good to go.

joelmuraguri: Alright, got it. I'm pushing the changes

PR #911, opened 2026-01-30
Implement account switching business logic

joelmuraguri:


data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -74,6 +75,10 @@ interface AuthRepository {
         request: SessionRequest,
     ): Outcome
 
+    suspend fun switchAccount(
```
tunjid: This method should be more restrictive. What happens if a `ProfileId` is passed that no session exists for?

It should take a `SessionSummary` instead.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -166,6 +171,26 @@ internal class AuthTokenRepository(
             onFailure = Outcome::Failure,
         )
 
+    override suspend fun switchAccount(
+        profileId: ProfileId,
+    ): Outcome = runCatchingUnlessCancelled {
```
tunjid, line 176: This method needs to be aware of the current session in. It needs to be cancelled if the session switching itself is cancelled.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -172,24 +171,32 @@ internal class AuthTokenRepository(
...
+    ): Outcome = savedStateDataSource.inCurrentProfileSession { signedInProfileId ->
+        if (sessionSummary.profileId == signedInProfileId) return@inCurrentProfileSession Outcome.Success
+
+        val authTokens = savedStateDataSource.savedState
+            .first()
+            .getAuthForProfile(sessionSummary.profileId)
+            ?: return@inCurrentProfileSession expiredSessionOutcome()
 
-        savedStateDataSource.setActiveProfile(profileId)
+        savedStateDataSource.setAuth(authTokens)
```
tunjid: These 3 methods suspend and happen one after the other, that will cause concurrency issues. The values should be set once and done atomically. This means `savedStateDataSource.updateSignedInProfileData` cannot be used at all.

Also note that setting a new active profile, will immediately cause the previous session to expire.

So you will need to first call `inCurrentProfileSession`, and when you get an expired session outcome, check if the new current session is the same as the one requested, and then change that expired outcome to a success one.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -76,7 +75,7 @@ interface AuthRepository {
     ): Outcome
 
     suspend fun switchAccount(
```
tunjid: I think `switchSession` would be a better name, since we're switching between past sessions.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
tunjid: You will probably need to add a dedicated abstract method on `SavedStateDataSource` that allows you to swap sessions using one call to it.

tunjid: Implementation in `DataStoreSavedStateDataSource`:

```kotlin
    suspend fun switchSession(
        profileId: ProfileId,
    ) = updateState {
        val profileData = profileData[profileId]
        val authTokens = profileData?.auth as? SavedState.AuthTokens.Authenticated
        if (profileData == null || authTokens == null) this
        else copy(activeProfileId = profileId)
    }
```

When you call this in `AuthRepository`, the `inCurrentProfileSession` block you are in will be cancelled immediately with an expired outcome session. This is good. You can then use a second `inCurrentProfileSession` block to check if the new session is the one you intended to switch to.

joelmuraguri: Refactored session switching to be atomic and removed non-atomic profile/auth updates. Previous session is expected to expire; the new session is validated in a follow-up `inCurrentProfileSession` block.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/SavedStateDataSource.kt
```diff
@@ -255,7 +255,7 @@ internal fun SavedState.signedProfilePreferencesOrDefault(): Preferences =
 internal fun SavedState.signedNotificationPreferencesOrDefault(): NotificationPreferences =
     signedInProfileData?.notifications?.preferences ?: NotificationPreferences.Default
 
-private fun SavedState.AuthTokens?.ifSignedIn(): SavedState.AuthTokens.Authenticated? =
+internal fun SavedState.AuthTokens?.ifSignedIn(): SavedState.AuthTokens.Authenticated? =
```
tunjid: This should remain private, it's an internal implementation detail in this file.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -166,6 +171,36 @@ internal class AuthTokenRepository(
...
+            if (signedInProfileId != sessionSummary.profileId) {
+                savedStateDataSource.switchSession(
+                    profileId = sessionSummary.profileId,
+                )
+            }
+            Outcome.Success
+        }
+
+        val currentProfileId =
+            savedStateDataSource.savedState.value.auth
```
tunjid: There are a semantics leaking here and values being read outside of the second `inCurrentProfileSession` lambda. The first `inCurrentProfileSession` should always return an expired session outcome. The `Outcome.Success` you return in it should not be seen in the success or failure case. You should assert that.

```kotlin
        val sessionSwitchOutcome = savedStateDataSource.inCurrentProfileSession { signedInProfileId ->
            if (signedInProfileId != sessionSummary.profileId) {
                savedStateDataSource.switchSession(
                    profileId = sessionSummary.profileId,
                )
            }
            Outcome.Success
        }

    if(sessionSwitchOutcome is Outcome.Success) return@runCatchingUnlessCancelled Outcome.Failure("Session switch failed")

        savedStateDataSource.inCurrentProfileSession { signedInProfileId ->
            if (signedInProfileId == sessionSummary.profileId) {
                updateSignedInUser()
            } else {
                expiredSessionOutcome()
            }
        } 
```

A valid profile id for a given session can only be read inside a `inCurrentProfileSession` block. Reading it outside and having coroutines suspend afterwards is a race condition.

tunjid: In the snippet above, we expect to not see the success, bc the starting session should be ended on a successful session switch.

After that, we should be in a new session. In that new session's context, evaluate that the switch actually completed. No logic should exist outside of a `inCurrentProfileSession`

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -166,6 +171,27 @@ internal class AuthTokenRepository(
...
+        sessionSummary: SessionSummary,
+    ): Outcome = runCatchingUnlessCancelled {
+        savedStateDataSource.inCurrentProfileSession { signedInProfileId ->
+            if (signedInProfileId != sessionSummary.profileId) {
+                savedStateDataSource.switchSession(
+                    profileId = sessionSummary.profileId,
+                )
+            }
+            expiredSessionOutcome()
+        }
```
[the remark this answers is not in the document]
tunjid: @joelmuraguri see what I mean about the value being discarded? If you don't intend to do anything with it, there's no need to return it.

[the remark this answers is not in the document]
joelmuraguri: Oh yes I see what you mean now. `expiredSessionOutcome()` is redundant here, should be deleted.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -166,6 +171,26 @@ internal class AuthTokenRepository(
...
+
+        savedStateDataSource.inCurrentProfileSession { signedInProfileId ->
+            if (signedInProfileId == sessionSummary.profileId) {
+                updateSignedInUser()
+            } else {
+                expiredSessionOutcome()
+            }
+        } ?: expiredSessionOutcome()
+    }.toOutcome()
```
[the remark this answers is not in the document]
tunjid: This will not account for a force logout from elsewhere. Ideally, the result of the switch should be used. This can be refactored later.

PR #936, opened 2026-02-02
Account switching ui implementation

joelmuraguri:


data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/UserDataRepository.kt
```diff
@@ -12,6 +13,8 @@ import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
 
 interface UserDataRepository {
+    val sessionSummaries: Flow<List<SessionSummary>>
```
tunjid: We already have `pastSession` in `AuthRepository`, no need to duplicate this here.

joelmuraguri: Yeah I see it, thanks for pointing that out.

feature/settings/src/commonMain/kotlin/com/tunjid/heron/settings/ui/AccountSwitchingItem.kt
```diff
@@ -0,0 +1,186 @@
...
+
+    Icon(
+        modifier = Modifier.graphicsLayer { rotationX = rotation },
+        imageVector = Icons.Default.ExpandLess,
+        contentDescription = stringResource(
+            if (isExpanded) CommonStrings.collapse_icon
+            else CommonStrings.expand_icon,
+        ),
+    )
+}
```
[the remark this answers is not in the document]
joelmuraguri: The suggested change intends to remove animation from `AccountSwitchingItem`, will ignore and  leave animations as is.

tunjid:
Sorry will review this soon. There's a lot of moving parts in the sign up flow with no tests. This change affects the sign in screen, I need to set time aside to review.

joelmuraguri:
@tunjid

feature/settings/src/commonMain/kotlin/com/tunjid/heron/settings/SettingsScreen.kt
```diff
@@ -43,13 +45,24 @@ internal fun SettingsScreen(
     actions: (Action) -> Unit,
     modifier: Modifier = Modifier,
 ) {
+    val enableItem = !state.isSwitchingAccount
```
tunjid: If a user is switching accounts, why is that state on the settings screen?

feature/settings/src/commonMain/kotlin/com/tunjid/heron/settings/SettingsScreen.kt
```diff
@@ -86,20 +102,24 @@ internal fun SettingsScreen(
...
         FeedbackItem(
             modifier = Modifier
                 .animateBounds(paneScaffoldState),
+            enabled = !state.isSwitchingAccount,
         )
         OpenSourceLibrariesItem(
             modifier = Modifier
                 .animateBounds(paneScaffoldState),
             libraries = state.openSourceLibraries,
+            enabled = enableItem,
```
tunjid: Rather than force each item to have an `enabledFlag`, it is better to just dim the entire screen.

feature/settings/src/commonMain/kotlin/com/tunjid/heron/settings/SettingsScreen.kt
joelmuraguri: This feature is for signedInProfiles, `SwitchSession` action is done in Settings screen.  This is borrowed from `isSubmitting` in `createSessionMutations`. The `isSwitchingAccount`state handles dimming out other action items in settings screen when `SwitchSession` action is clicked and triggers indicator for better UI experience instead of user experiencing only delay when action is clicked.

tunjid (commented):
A few things:

* A video recording of the UI implemented would be really helpful.
* I don't think disabling individual settings items is the way to go for this. If switching accounts will take time, I'd much rather a dimming of the UI and a progress bar blocking the user from leaving with the switch takes place occurs.

I suggest reverting the changes to the individual items, and adding a "Switching accounts..." dialog or something similar. You can get creative with an animation in a separate PR if you want.

It is also possible the account switched to has an expired token. You should consider how to recover from this.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -175,7 +177,28 @@ internal class AuthTokenRepository(
     override suspend fun switchSession(
         sessionSummary: SessionSummary,
     ): Outcome = runCatchingUnlessCancelled {
-        // Switching should cause the current session to expire
+        // Get target token from SavedState
+        val currentState = savedStateDataSource.savedState.first()
+        val versionedState = currentState as? VersionedSavedState
```
tunjid: This cast to the explicit type is a code smell. Why do you need the actual type?

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -175,7 +177,28 @@ internal class AuthTokenRepository(
...
+            ?: return@runCatchingUnlessCancelled Outcome.Failure(
+                SessionSwitchException(sessionSummary.profileId),
+            )
+
+        val targetProfileData = versionedState.profileData[sessionSummary.profileId]
```
tunjid: This does not seem thread or session safe.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -337,4 +373,24 @@ internal class AuthTokenRepository(
...
+        authTokens: SavedState.AuthTokens.Authenticated,
+    ) = runCatchingUnlessCancelled {
+        sessionManager.refreshSessionToken(authTokens)
+    }.fold(
+        onSuccess = { refreshedAuth ->
+            savedStateDataSource.setAuth(refreshedAuth)
+
+            // Wait for token to be saved
+            savedStateDataSource.savedState.first { state ->
+                val currentAuth = state.auth as? SavedState.AuthTokens.Authenticated
```
tunjid: Should this have a timeout?

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
joelmuraguri: I see that this is architecturally wrong. There's no need for the actual type here. We should use `SavedStateDataSource` for read and write operations instead of  `VersionedSavedState` concrete implementation. Instead I think an extension helper will handle this well.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
joelmuraguri: Yes we should have it.  If write fails, dispatcher is cancelled and auth is invalidated. This can suspend forever. 
Thanks for pointing that out.

joelmuraguri: I think this is method and its implementaion  is unnecessary if we update target profile `SavedState` atomically. This will remove race conditions.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
joelmuraguri: Good catch, this was indeed not session-safe since the target auth state was being read outside a guarded session context.

I’ve updated the implementation so that Token refresh is performed within the `inCurrentProfileSession` guarded context before the switch.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -175,12 +175,26 @@ internal class AuthTokenRepository(
...
+        val freshAuth = savedStateDataSource.inCurrentProfileSession { _ ->
+            val targetProfileData = savedStateDataSource.savedState.value
+                .profileData[sessionSummary.profileId]
+                ?: return@inCurrentProfileSession null
+
+            val targetAuthTokens =
+                targetProfileData.auth as? SavedState.AuthTokens.Authenticated
+                    ?: return@inCurrentProfileSession null
+
+            sessionManager.refreshSessionToken(targetAuthTokens)
```
tunjid: Let's take a step back and think about what we're trying to achieve, and how what you've written works.

At any one time, there can be several network request happening concurrently. When switching a session, it will happen while other network requests are in flight.

What happens if you try to refresh the token of another profile with the same http client the current profile is using to make requests?

You are brute forcing this solution. You should take a step back, and think about what you're trying to do, the constraints of the system and utilize the tools for working in concurrent environments: Atomicity, isolation and so on.

tunjid: For example, at the very least each profile should have its own http client in the session manager. You may wan't to think about how the session manager architecture might need to evolve to accommodate this.

joelmuraguri: Thanks for the detailed feedback, this makes sense.

I'm going to step back and rethink this solution and session manager architecture with concurrency concepts in mind, then come back with a clearer approach instead of brute forcing.

Appreciate you calling this out.

joelmuraguri:
> A few things:
> 
> * A video recording of the UI implemented would be really helpful.
> * I don't think disabling individual settings items is the way to go for this. If switching accounts will take time, I'd much rather a dimming of the UI and a progress bar blocking the user from leaving with the switch takes place occurs.
> 
> I suggest reverting the changes to the individual items, and adding a "Switching accounts..." dialog or something similar. You can get creative with an animation in a separate PR if you want.
> 
> It is also possible the account switched to has an expired token. You should consider how to recover from this.

I’m planning a follow-up PR to add the dimmed blocking state, progress indication and animations. I’ll also include a video recording of the updated UI.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -177,21 +179,54 @@ internal class AuthTokenRepository(
     override suspend fun switchSession(
         sessionSummary: SessionSummary,
     ): Outcome = runCatchingUnlessCancelled {
-        // Switching should cause the current session to expire
+        // Get the stored auth for the target session
```
tunjid: Shouldn't this be done in a `inCurrentProfileSession` block?

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/SavedStateDataSource.kt
```diff
@@ -264,6 +267,12 @@ private fun SavedState.AuthTokens?.ifSignedIn(): SavedState.AuthTokens.Authentic
         -> null
     }
 
+internal val SavedState.profileData: Map<ProfileId, SavedState.ProfileData>
```
tunjid: This feels hacky. You should instead create a `inPastSession(pastSession) {}` block that uses with context to set the session context instead of exposing this. You may also want to add `ProfileData` to a session context.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -177,21 +179,54 @@ internal class AuthTokenRepository(
...
+        val targetProfileAuth = savedStateDataSource.savedState.value
```
tunjid: I think you want to create a `inPastSession` extension on `SavedStateDataSource`, and then use that here. That same logic will be useful when receiving notifications from other profiles other than the signed in one.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/AuthRepository.kt
```diff
@@ -177,23 +179,37 @@ internal class AuthTokenRepository(
...
-        val switched = savedStateDataSource.inCurrentProfileSession { signedInProfileId ->
-            if (signedInProfileId == sessionSummary.profileId) return@inCurrentProfileSession true
+        val switched = savedStateDataSource.inCurrentProfileSession { currentProfileId ->
+            val freshAuth = savedStateDataSource.inPastSession(sessionSummary.profileId) {
+                networkService.runCatchingWithMonitoredNetworkRetry {
+                    getSession()
+                }.getOrElse {
+                    return@inPastSession null
+                }
+                currentCoroutineContext()[SessionContext]?.tokens as? SavedState.AuthTokens.Authenticated
```
tunjid: Why do you need to read the session context here? Can you check if the value of the `getSession` call?

joelmuraguri: I’m reading from `SessionContext` here because the goal isn’t just to validate the session via `getSession()`, but to obtain the authenticated tokens needed for an atomic session switch.

`getSession()` only returns session metadata and doesn’t expose tokens. Since `inPastSession()` installs `SessionContext.Previous` with the authenticated tokens for that profile, the coroutine context is the reliable place to retrieve them during this flow.

tunjid: okie, I put some time on your cal for tomorrow where we can review this PR together. Should be easier then.

In the meantime, please rebase this on the current main

joelmuraguri: Got it. I’ve rebased on the current main. Looking forward to reviewing it together tomorrow.

Issue #670, opened 2025-11-21
[Feature Request] Publish app on F-droid/IzzyOnDroid or Github releases


kryoqa:
**Is your feature request related to a problem? Please describe.**
Having the app on F-droid or IzzyOnDroid would open up the app to more people who might be using custom ROMs. Even an apk release on Github would be appreciated

kryoqa:
The options presented when creating an issue - Feature Request, Blank, Bug report - is it working properly ? I edited the title to Feature request.

tunjid:
Hi! I'm open to this, I just don't know how.

kryoqa:
[F-droid guide](https://f-droid.org/en/docs/Submitting_to_F-Droid_Quick_Start_Guide/)

If f-droid feels like too much work, an apk on github releases would be fine 😅

tunjid:
I'll look into this! Just have a bunch of moderation features I'm trying to finish.

Olaren15:
+1 Having github apk releases to be able to install with obtainium would be greatly appreciated!

tunjid:
@kryoqa @Olaren15 I now have github actions for publishing to the play store. Should be easy to also push the artifact to github. Lemme look for resources for that.

Pandooroo:
+1 for a non-Play Store way to download Heron, especially for someone who wants to avoid Google services for life.

CompeyDev:
At the least, can we have GitHub releases? That way, tools like Obtainium can be used in the meantime.

It is relatively simple to setup an action to just compile and publish releases. I can even open a PR if you would be open to that.

For IzzyOnDroid, CC @IzzySoft, how should the maintainer proceed?

tunjid:
@CompeyDev yes please, do open a PR.

IzzySoft:
We'd also need the APKs attached to releases. Once everything is in place, be welcome opening an App Inclusion Request ~[at our issue tracker](https://gitlab.com/IzzyOnDroid/repo/-/issues)~ (which is our last repo at GitLab, and will hopefully soon move to our [Codeberg presence](https://codeberg.org/IzzyOnDroid), too – to our [repodata](https://codeberg.org/IzzyOnDroid/repodata) repo, we're just preparing labels, issue templates, etc for that).

**Edit:** Please file it [at this issue tracker](https://codeberg.org/IzzyOnDroid/repodata/issues) then, as we're in the process to just cleaning up the original one.

tunjid:
[...]

PR #960, opened 2026-02-06
Publish APKs to GitHub releases in CI

CompeyDev:
Self-explanatory title. Relates to #670.

Not sure how to test it honestly, looks like the workflow triggers depend on PR merges containing a title of a specific format. How do you suggest I test these changes?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -72,3 +72,23 @@ jobs:
           track: internal
           mappingFile: composeApp/build/outputs/mapping/release/mapping.txt
           debugSymbols: composeApp/build/intermediates/merged_native_libs/release/mergeReleaseNativeLibs/out/lib
+
+
+      - name: Extract signed APKs
+        run: |
+          curl -L -o bundletool.jar https://github.com/google/bundletool/releases/download/1.18.3/bundletool-all-1.18.3.jar
+          java -jar bundletool.jar build-apks \
+            --bundle=${{ steps.sign_app.outputs.signedReleaseFile }} \
+            --output=composeApp/build/outputs/apk/release/heron.apks \
+            --mode=universal
+
+          unzip -o composeApp/build/outputs/apk/release/heron.apks -d composeApp/build/outputs/apk/release/
+
+      - name: Upload to GitHub releases
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: v${{ github.run_number }}
+          name: Heron v${{ github.run_number }}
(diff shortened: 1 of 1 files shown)


tunjid:
Yes it runs each time I want to cut a release for Google play. The app doesn't have any incomplete features exposed atm, so I'll cut a release soon.

.github/workflows/publish.yml
```diff
@@ -72,3 +72,23 @@ jobs:
...
+            --mode=universal
+
+          unzip -o composeApp/build/outputs/apk/release/heron.apks -d composeApp/build/outputs/apk/release/
+
+      - name: Upload to GitHub releases
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: v${{ github.run_number }}
+          name: Heron v${{ github.run_number }}
+          files: composeApp/build/outputs/apk/release/universal.apk
```
tunjid, line 92: I think this will need to be the signed release file from the step above: `${{steps.sign_app.outputs.signedReleaseFile}}`

Heron has Google Play signing on, so the github releases will have a different key than the one ultimately published to Google Play. In hindsight, I probs shouldn't have opted in.

tunjid: Unless the raw apk is the desired upload.

CompeyDev: No, the file that we are referring to is what is emitted by `bundletool` *after* we extract the AAB into APKs. That expression would yield the AAB file, which we use in the previous step.

CompeyDev:
> Yes it runs each time I want to cut a release for Google play. The app doesn't have any incomplete features exposed atm, so I'll cut a release soon.

That would make it challenging for me to figure out whether the workflow works though. How do we test it?

.github/workflows/publish.yml
tunjid: Ah I see, ty.

tunjid:
> > Yes it runs each time I want to cut a release for Google play. The app doesn't have any incomplete features exposed atm, so I'll cut a release soon.
> 
> That would make it challenging for me to figure out whether the workflow works though. How do we test it?

I can just merge this as is and run a build of the current main. The Google Play uploads go to the internal testing track, I don't have to promote to production.

tunjid:
@CompeyDev I can also add you as a collaborator to the project so you can run actions on the main repo and this branch instead of the fork?

CompeyDev:
No, that's fine. You can merge this as-is and try debugging the action when you publish a release if that is okay with you. Good luck!

PR #984, opened 2026-02-15
Session switching animations & shared element transition

joelmuraguri:
## DEMO

https://github.com/user-attachments/assets/814c8a8a-ff53-4202-b061-c27ce222e9cc


tunjid:
Oh this is so tastefully done! We'll review along with the main PR today.

PR #1012, opened 2026-02-19
Support FeedLists in ThreadGate

joelmuraguri:


feature/moderation/src/commonMain/kotlin/com/tunjid/heron/moderation/ModerationScreen.kt
```diff
@@ -83,6 +83,8 @@ internal fun ModerationScreen(
     modifier: Modifier = Modifier,
 ) {
     val threadGateSheetState = rememberUpdatedThreadGateSheetState(
+        recentLists = emptyList(),
```
tunjid: These need to be populated, otherwise the user can's use them.

feature/compose/src/commonMain/kotlin/com/tunjid/heron/compose/ui/ComposeThreadGate.kt
```diff
@@ -33,6 +33,8 @@ fun ComposeThreadGate(
     onInteractionSettingsUpdated: (PostInteractionSettingsPreference) -> Unit,
 ) {
     val threadGateSheetState = rememberUpdatedThreadGateSheetState(
+        recentLists = emptyList(),
```
tunjid: This also should be populated.

tunjid (approved):

PR #1036, opened 2026-02-23
Expose parentPostUri in ThreadPostEntity

joelmuraguri:
Adds `parentPostUri` to `ThreadedPostEntity` and updates the `postThread` DAO query to select it via a LEFT JOIN on the postThreads table.

### DIFF
--- a/data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/daos/PostDao.kt
+++ b/data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/daos/PostDao.kt
@@ -397,7 +397,7 @@ interface PostDao {
               ft.generation AS generation,
               ft.ancestorCreated AS ancestorCreated,
               ft.postCreated AS postCreated,
-              CASE WHEN ft.generation = 0 THEN NULL ELSE pt.parentPostUri END AS parentPostUri
+              pt.parentPostUri AS parentPostUri
             FROM
               posts p
             JOIN


data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/PostEntity.kt
```diff
@@ -206,6 +206,7 @@ data class ThreadedPostEntity(
     val entity: PostEntity,
     val generation: Long,
     val rootPostUri: PostUri?,
+    val parentPostUri: PostUri?,
```
tunjid, line 209: You can't update an entity without a migration, the app will crash.

tunjid: Oh I see, this is just an Entity in a query, it's not a persisted type.

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/daos/PostDao.kt
```diff
@@ -396,11 +396,14 @@ interface PostDao {
               ft.rootPostUri AS rootPostUri,
               ft.generation AS generation,
               ft.ancestorCreated AS ancestorCreated,
-              ft.postCreated AS postCreated
+              ft.postCreated AS postCreated,
+              CASE WHEN ft.generation = 0 THEN NULL ELSE pt.parentPostUri END AS parentPostUri
```
tunjid: Lets remove the `CASE WHEN` here and just have `pt.parentPostUri AS parentPostUri`. The engine will be able to optimize the query better

PR #1044, opened 2026-02-25
Impl ReplyTree data layer

joelmuraguri:


data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/TimelineRepository.kt
```diff
@@ -1173,59 +1174,94 @@ internal class OfflineTimelineRepository(
     ) = with(context) {
         val lastItem = list.lastOrNull()
         when {
-            // To start or for the OP, start a new thread
```
tunjid, line 1176: I think this method should be left alone, its trying to do two things. You should have a different method for reply building.

joelmuraguri: After switching `postThreadedItems` to use `::spinReplyTree`, `spinThread` is no longer called anywhere in the codebase. I want to make sure I understand the intent here, should spinThread be removed for now and reintroduced when there is an actual call site for it or is there a planned use case coming that I should be aware of so I can keep it in place?

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/state/Timeline.kt
```diff
@@ -247,6 +247,7 @@ private fun TiledList<TimelineQuery, TimelineItem>.filterThreadDuplicates(): Til
 
             is TimelineItem.Single -> !threadRootIds.contains(item.post.cid)
             is TimelineItem.Placeholder -> false
+            is TimelineItem.ReplyTree -> false
```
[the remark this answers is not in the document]
joelmuraguri: `ReplyTree` is only produced by `postThreadedItems` for `PostDetailScreen` via a plain List, never a `TiledList`. It cannot reach `filterThreadDuplicates`. Returning false is the correct defensive fallback, grouped with `Placeholder` for the same reason.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/TimelineRepository.kt
joelmuraguri: Oh, I see, why we need the two methods as separate

Both methods are now active. `spinThread` handles `ReplyViewMode.Linear` and `spinReplyTree` handles `ReplyViewMode.Threaded`. The user can toggle between them.

PR #1046, opened 2026-02-26
Render reply tree on post detail screen

joelmuraguri:
## Updated Screen Demo

https://github.com/user-attachments/assets/cb7847cd-004a-4a0f-b346-7d6b16f59677


tunjid:
I think the UI needs quite a bit of work. The gaps in the timeline lines in the UI can't be merged in this state. If there's any indentation in the reply node, there should be a curved arrow to the reply like in the bsky app.

Also I don't think this is shippable without a toggle between the old and new behavior

PR #1049, opened 2026-02-27
Support Profile Live Status, Data Layer

joelmuraguri:


data/database/schemas/com.tunjid.heron.data.database.AppDatabase/34.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 34,
-    "identityHash": "1e975be4633a2c929f918f6201d3bdf4",
```
tunjid, line 5: If the DB identity hash changes, it means you're missing a migration. The identity has hash for a DB version shouldn't change.

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/ProfileStatusEntity.kt
```diff
@@ -0,0 +1,59 @@
+package com.tunjid.heron.data.database.entities
+
+import androidx.room.Entity
+import androidx.room.ForeignKey
+import com.tunjid.heron.data.core.models.Profile
+import com.tunjid.heron.data.core.types.ImageUri
+import com.tunjid.heron.data.core.types.ProfileId
+import kotlin.time.Instant
+
+@Entity(
```
tunjid: I don't think creating a `ProfileStatusEntity` is needed. A profile can only have one status, and that status is the one with the `self` record key. Each time a profile goes live, its always updating its self status. 

That is, a profile status is a 1:1 record. The status should be a nullable embedded record in the `ProfileEntity`

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/Profile.kt
```diff
@@ -114,6 +115,49 @@ data class Profile(
         val chat: ChatInfo,
     )
 
+    @Serializable
+    data class ProfileStatus(
```
tunjid, line 119: Since this is most likely going to be an embedded model in `ProfileEntity`, we should make it a flat data class.

tunjid: i.e:

```kotlin
    data class ProfileStatus(
        val uri: String? = null,
        val status: String,
        val uri: String,
        val title: String,
        val description: String,
        val thumb: ImageUri? = null,
        val expiresAt: Instant? = null,
        val isActive: Boolean? = null,
        val isDisabled: Boolean? = null,
    )
```

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/ProfileEntity.kt
```diff
@@ -56,6 +56,8 @@ data class ProfileEntity(
     val createdAt: Instant?,
     @Embedded
     val associated: Associated,
+    @Embedded(prefix = "live_")
```
tunjid: Would `status_` be a better prefix?

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/migrations/Migration34To35ProfileStatusAutoMigration.kt
```diff
@@ -0,0 +1,19 @@
+package com.tunjid.heron.data.database.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.SQLiteConnection
+import androidx.sqlite.execSQL
+
+internal object Migration34To35ProfileStatus : Migration(34, 35) {
```
tunjid: Since this just adds nullable columns, I think an auto migration would be easier: https://medium.com/androiddevelopers/room-auto-migrations-d5370b0ca6eb

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
```diff
@@ -115,6 +121,11 @@ interface ProfileRepository {
         limit: Long,
     ): Flow<List<Profile>>
 
+    fun listMembers(
```
tunjid: This seems like a rebase error, this method is no longer on this interface.

data/database/schemas/com.tunjid.heron.data.database.AppDatabase/34.json
```diff
@@ -1041,6 +1041,78 @@
           }
         ]
       },
+      {
```
tunjid: I believe changes to `34.json` need to be reverted

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/ProfileEntity.kt
joelmuraguri: Yes this is definitive and matches naming for this feature, **Profile Live Status**

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/migrations/Migration34To35ProfileStatusAutoMigration.kt
joelmuraguri: Thanks for the article!

PR #1061, opened 2026-03-03
Profile live status, UI layer

joelmuraguri:
## Demo

https://github.com/user-attachments/assets/a6562963-14d8-42d3-9e23-38a309848453

### DIFF
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
@@ -475,7 +475,7 @@ internal class OfflineProfileRepository @Inject constructor(
             is Profile.StatusUpdate.GoLive -> networkService.runCatchingWithMonitoredNetworkRetry {
                 putRecord(
                     PutRecordRequest(
-                        repo = update.profileId.id.let(::Did),
+                        repo = update.signedInProfileId.id.let(::Did),
                         collection = Nsid(Collections.ProfileStatus),
                         rkey = RKey(Collections.SelfRecordKey.rkey),
                         record = Status(
@@ -498,7 +498,7 @@ internal class OfflineProfileRepository @Inject constructor(
                 val now = Clock.System.now()
                 val expiresAt = now.plus(update.durationMinutes.minutes)
                 profileDao.updateStatus(
-                    did = update.profileId,
+                    did = update.signedInProfileId,
                     uri = null,
                     value = Profile.ProfileStatus.STATUS_LIVE,
                     uriLink = update.streamUrl,
@@ -514,14 +514,14 @@ internal class OfflineProfileRepository @Inject constructor(
             is Profile.StatusUpdate.EndLive -> networkService.runCatchingWithMonitoredNetworkRetry {
                 deleteRecord(
                     DeleteRecordRequest(
-                        repo = update.profileId.id.let(::Did),
+                        repo = update.signedInProfileId.id.let(::Did),
                         collection = Nsid(Collections.ProfileStatus),
                         rkey = RKey(Collections.SelfRecordKey.rkey),
                     ),
                 )
             }.toOutcome {
                 profileDao.updateStatus(
-                    did = update.profileId,
+                    did = update.signedInProfileId,
                     uri = null,
                     value = null,
                     uriLink = null,
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/writequeue/Writable.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/writequeue/Writable.kt
@@ -151,7 +151,7 @@ sealed interface Writable {
     ) : Writable {
 
         override val queueId: String
-            get() = "status-update-${update.profileId}"
+            get() = "status-update-${update.signedInProfileId}"
 
         override suspend fun WriteQueue.write(): Outcome =
             profileRepository.updateProfileStatus(update)
--- a/data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/Profile.kt
+++ b/data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/Profile.kt
@@ -138,18 +138,18 @@ data class Profile(
 
     @Serializable
     sealed class StatusUpdate {
-        abstract val profileId: ProfileId
+        abstract val signedInProfileId: ProfileId
 
         @Serializable
         data class GoLive(
-            override val profileId: ProfileId,
+            override val signedInProfileId: ProfileId,
             val streamUrl: String,
             val durationMinutes: Int = ProfileStatus.DEFAULT_LIVE_DURATION_MINUTES,
         ) : StatusUpdate()
 
         @Serializable
         data class EndLive(
-            override val profileId: ProfileId,
+            override val signedInProfileId: ProfileId,
         ) : StatusUpdate()
     }
 


feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ui/ProfileActionsMenu.kt
```diff
@@ -141,8 +144,24 @@ internal sealed class ProfileActionMenu {
     data object Divider : ProfileActionMenu()
 }
 
-internal fun ProfileViewerState?.profileActionMenuItems() = buildList {
-    if (this@profileActionMenuItems != null) {
+internal fun ProfileViewerState?.profileActionMenuItems(
+    isSignedInProfile: Boolean = false,
```
tunjid: Let's remove the default arguments. We should always explicitly pass something

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ui/ProfileActionsMenu.kt
```diff
@@ -141,8 +144,24 @@ internal sealed class ProfileActionMenu {
...
+                    title = CommonStrings.action_edit_live_status,
+                    icon = Icons.Rounded.Videocam,
+                )
+                else -> ProfileActionMenu.Item(
+                    title = CommonStrings.action_go_live,
+                    icon = Icons.Rounded.Videocam,
+                )
+            },
+        )
+    } else if (this@profileActionMenuItems != null) {
```
tunjid: Do we want to use an `else if` here? It seems independent from whether or not it is the signed in profile. What happens if it's still just `if (this@profileActionMenuItems != null)`?

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
```diff
@@ -809,20 +841,24 @@ private fun ProfileAvatar(
...
-                rememberSharedContentState(
-                    key = avatarSharedElementKey,
-                )
+                rememberSharedContentState(key = avatarSharedElementKey)
             },
             zIndexInOverlay = AvatarZIndex,
             modifier = Modifier
                 .fillMaxSize()
                 .padding(headerState.avatarPadding)
+                .then(
```
tunjid: We have a nice extension for this. It can be:

```
.ifTrue(isLive) {
    border(
         width = LiveBorderWidth,
         color = LiveStatusColor,
         shape = CircleShape,
     ) 
}
```

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/State.kt
```diff
@@ -200,6 +200,18 @@ sealed class Action(val key: String) {
         val target: LinkTarget,
     ) : Action(key = "BioLinkClicked")
 
+    sealed class LiveStatus(key: String) : Action(key) {
```
tunjid: Lets make this a verb with `UpdateLiveStatus`

scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/scaffold/PaneTopAppBar.kt
```diff
@@ -97,31 +103,44 @@ fun PaneScaffoldState.RootDestinationTopAppBar(
...
                     ) {
-                        AsyncImage(
+                        PaneStickySharedElement(
                             modifier = Modifier
-                                .fillParentAxisIfFixedOrWrap()
-                                .clickable {
-                                    onSignedInProfileClicked(
-                                        profile,
-                                        UiTokens.SignedInUserAvatarSharedElementKey,
+                                .fillMaxSize()
```
tunjid: The border `Modifier` should be on the `AsyncImage`, not the `PaneStickySharedElement`, otherwise it won't be part of the shared element transition.

ui/core/src/commonMain/kotlin/com/tunjid/heron/ui/LiveChip.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.compose.ui.unit.sp
+import com.tunjid.heron.ui.text.CommonStrings
+import heron.ui.core.generated.resources.live_badge
+import org.jetbrains.compose.resources.stringResource
+
+@Composable
+fun LiveChip(modifier: Modifier = Modifier) {
+    Box(
+        modifier = modifier
+            .offset(y = 5.dp)
```
tunjid: Do we need to specify the offset here? I think the parent Composable should set it, or use `Modifier.align()` if the parent Composable is a `Box`

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ui/ProfileActionsMenu.kt
joelmuraguri: Reverted, tested with a plain if and block/mute showed up on your own profile. The API returns a non-null viewerState even for self, so else if is needed to keep the two branches mutually exclusive.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
```diff
@@ -139,6 +140,11 @@ interface ProfileRepository {
     suspend fun updateProfileStatus(
         update: Profile.StatusUpdate,
     ): Outcome
+
+    suspend fun refreshProfile(
```
tunjid: Why do you need this? `ProfileRepository.profile()` exists, which does the same thing if you do `ProfileRepository.profile().take(1).collect()`. Though I'm not sure why you'd need that either.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
```diff
@@ -475,7 +481,7 @@ internal class OfflineProfileRepository @Inject constructor(
             is Profile.StatusUpdate.GoLive -> networkService.runCatchingWithMonitoredNetworkRetry {
                 putRecord(
                     PutRecordRequest(
-                        repo = update.profileId.id.let(::Did),
+                        repo = update.signedInProfileId.id.let(::Did),
```
tunjid: Just a note, and not actionable as an update hasn't gone out yet. When an update goes out, these values will be used and serialized making it hard to change variable names without migration.

The app uses:

* Cbor for serializing url params: These are sensitive to field names.
* Protobuf for serializing to datastore: These are sensitive to field order.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
joelmuraguri: We need it to refresh Profile model once live status is updated. `ProfileStatus` could be null, when we `updateLiveStatus`. We need to fetch from network and update `ProfileEntity` otherwise, the root app bar  and other UI parts of the application won't display Live status badge. This is a fix from earlier which didn't persist a ProfileLiveStatus across the app.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
```diff
@@ -958,29 +998,35 @@ private fun ProfileHeadline(
...
                                     onItemClicked = { item ->
                                         when (item.title) {
+                                            CommonStrings.action_go_live,
+                                            CommonStrings.action_edit_live_status,
+                                            -> onUpdateProfileLiveStatus()
                                             CommonStrings.viewer_state_block_account ->
                                                 profileRestrictionsDialogState.show(
                                                     Action.Block.Add(
-                                                        signedInProfileId = signedInProfileId,
+                                                        signedInProfileId = signedInProfileId!!,
```
tunjid: I'm not fond of double bangs. I think it's better to be explicit and use `requireNotNull(signedInProfileId)`, or use an `if` statement to check its not null.

In this case, as its UI code and not data layer code where exceptions are caught, an `if` statement is safer.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
```diff
@@ -958,29 +998,35 @@ private fun ProfileHeadline(
                                     else onViewerStateClicked(viewerState)
                                 },
                             )
-                            if (!isSignedInProfile && signedInProfileId != null) {
+                            if (isSignedInProfile || signedInProfileId != null) {
```
tunjid: To address the comment above, what if this was:

```
if (signedInProfileId != null && isSignedInProfile) {
    ...
}
```

I believe the logic is the same? You also get a null check.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileViewModel.kt
```diff
@@ -452,6 +459,60 @@ private fun Flow<Action.SendPostInteraction>.postInteractionMutations(
...
+        signedInProfileId = signedInProfileId,
+        profileRepository = profileRepository,
+    )
+}
+
+private suspend fun FlowCollector<Mutation<State>>.refreshLiveStatusMutations(
+    signedInProfileId: ProfileId,
+    profileRepository: ProfileRepository,
+) {
+    emit { copy(signedInProfileId = signedInProfileId) }
```
tunjid: Why do you need to manually update the `signedInProfileId`? It's already being updated by `loadProfileMutations`.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
joelmuraguri: Oh Yes this will remove the non-null assertion

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileViewModel.kt
```diff
@@ -452,6 +459,60 @@ private fun Flow<Action.SendPostInteraction>.postInteractionMutations(
...
+                signedInProfileId = action.signedInProfileId,
+            )
+        },
+    )
+    val status = writeQueue.enqueue(writable)
+    writable.writeStatusMessage(status)?.let {
+        emit { copy(messages = messages + it) }
+    }
+
+    refreshLiveStatusMutations(
```
tunjid: If performing a status update updates a profile, it should be done in the data layer after the write. It should not depend on an independent call to refresh the profile.

You already call `profileDao.updateStatus`, why do you need an explicit call to refresh the profile?

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileViewModel.kt
```diff
@@ -531,21 +592,6 @@ private fun CoroutineScope.recordStateHolders(
                 cursorListLoader = recordRepository::feedGenerators,
             ),
         ),
-        ProfileScreenStateHolders.Records.Documents(
```
tunjid, line 534: Did you mean to delete this? Looks like a rebase error.

joelmuraguri: No that wasn't my intention

scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/scaffold/PaneTopAppBar.kt
```diff
@@ -97,31 +104,44 @@ fun PaneScaffoldState.RootDestinationTopAppBar(
...
-                            .size(36.dp),
-                        sharedContentState = rememberSharedContentState(
-                            key = UiTokens.SignedInUserAvatarSharedElementKey,
-                        ),
+                    Box(
+                        modifier = Modifier.size(36.dp),
+                        contentAlignment = Alignment.BottomCenter,
                     ) {
-                        AsyncImage(
+                        PaneStickySharedElement(
```
tunjid, line 105: The live chip needs to be part of the shared element, otherwise you'll get flashing when navigating between tabs. `PaneStickySharedElement` should be the parent, and the `Box`, `AsyncImage,` and `LiveChip` should be its children.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileViewModel.kt
joelmuraguri: This was first implementation but I noticed, the `ProfileLiveStatus` isnt  getting persisted. The change is immediately seen on ProfileScreen but when navigating across the app, the liveBadge isnt there anywhere in the app also the `ProfileActionMenu` still shows there is no live status. Clearly the Profile is live, confirmed this in Bluesky mobile app and web. 

I think the bug made me overthink all this.

joelmuraguri: ## Demo of what we previously had 


https://github.com/user-attachments/assets/3da1d023-0aeb-4198-a65f-ad6bf47e92ab

Once user exits the Profile Screen,  we can’t read the liveStatus again in app. 

I thought refreshing Profile would solve this but it seems it needs redesigning.

tunjid: This may have been bc of the way you're calling the `PaneSharedElement` API

PR #1114, opened 2026-03-17
Build out tree view data layer

joelmuraguri:
This PR is split into two commits to make review easier:
1. Define Threaded as a sealed class
`TimelineItem.Threaded`  promote to sealed class with `Linear` and `Tree` subtypes

`Threaded` was a single data class. It is now a sealed class with two subtypes:
- `Threaded.Linear` the previous Threaded/Thread shape, used in feed and tiled lists
- `Threaded.Tree` for post detail reply trees, with `Node` nested inside it

All call sites updated to reference `Threaded.Linear` and exhaustive when branches updated to include `Threaded.Tree`.

2. Construction logic for Tree view

- Adds `spinReplyTree` to construct a `Threaded.Tree` from threaded post entities, with depth and sibling caps to prevent OOM on viral posts
- Updates `replyNodeIndex` in `TimelineItemCreationContext` to `MutableMap<PostUri, MutableList<Threaded.Tree.Node>>`
- Wires spinReplyTree into `postThreadMutations` which will need update in `PostDetailViewModel` and `GalleryViewModel`

### DIFF
--- a/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/TimelineItem.kt
+++ b/ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/TimelineItem.kt
@@ -481,7 +481,7 @@ private val TimelineItem.isThreadedAncestor
     }
 
 private val TimelineItem.isThreadedAnchor
-    get() = this is TimelineItem.Threaded.Linear && generation == 0L
+    get() = this is TimelineItem.Threaded.Tree || (this is TimelineItem.Threaded.Linear && generation == 0L)
 
 private val TimelineItem.isThreadedAncestorOrAnchor
     get() = isThreadedAncestor || isThreadedAnchor


data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/TimelineRepository.kt
```diff
@@ -1234,6 +1239,96 @@ internal class OfflineTimelineRepository(
...
+                        post = post,
+                        ancestors = emptyList(),
+                        replies = emptyList(),
+                        isMuted = isMuted(post),
+                        threadGate = threadGate(post.uri),
+                        appliedLabels = appliedLabels,
+                        signedInProfileId = signedInProfileId,
+                    )
+                }
+            }
```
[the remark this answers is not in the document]
tunjid: We can ignore this.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/recordResolver/MutableTimelineItemCreationContext.kt
```diff
@@ -54,6 +54,7 @@ internal class MutableTimelineItemCreationContext(
         labelers = emptyList(),
         contentLabelPreferences = emptyList(),
     )
+    override val replyNodeIndex: MutableMap<PostUri, MutableList<TimelineItem.Threaded.Tree.Node>> = mutableMapOf()
```
tunjid, line 57: I don't think we need to add this to `MutableTimelineItemCreationContext`. Can we just create the map in the `spinReplyTree` method? We'd be leaking an abstraction with this, we only use it in one place.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/TimelineItem.kt
```diff
@@ -475,13 +475,13 @@ fun String.withQuotingPostUriPrefix(
     ?: this
 
 private val TimelineItem.isThreadedAncestor
-    get() = this is TimelineItem.Thread && when (val gen = generation) {
+    get() = this is TimelineItem.Threaded.Linear && when (val gen = generation) {
```
tunjid, line 478: Does this only apply to linear threads?

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/TimelineItem.kt
```diff
@@ -475,13 +475,13 @@ fun String.withQuotingPostUriPrefix(
...
         null -> false
         else -> gen <= -1
     }
 
 private val TimelineItem.isThreadedAnchor
-    get() = this is TimelineItem.Thread && generation == 0L
+    get() = this is TimelineItem.Threaded.Linear && generation == 0L
```
tunjid: Same for this, is this only for linear threads? I'm not sure if it is, just clarifying

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/recordResolver/MutableTimelineItemCreationContext.kt
joelmuraguri: `spinReplyTree` is called once per entity, so initializing the map inside it resets on every call descendants would never find their parent's bucket. The map must survive the entire tree-building pass, which is what the closure in `replyTreeSpinner` guarantees.

This is what I think is an alternative:

```
private fun replyTreeSpinner(): RecordResolver.TimelineItemCreationContext.(ThreadedPostEntity) -> Unit {
    val replyNodeIndex = mutableMapOf<PostUri, MutableList<TimelineItem.Threaded.Tree.Node>>()
    return { thread -> spinReplyTree(this, thread, replyNodeIndex) }
}
```

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/TimelineItem.kt
joelmuraguri: No I don't think so

`generation` property only exists on `Threaded.Linear`, `Threaded.Tree` expresses the same concept structurally via `ancestors` and `replies`.  So yes, I think I have scoped it correctly

joelmuraguri: Oh let me correct my logic here

`Threaded.Tree` doesn't have `generation` so it can't participate in  `isThreadedAncestor`, but it always qualifies for `isThreadedAnchor` since it's structurally the anchor by definition which is what drives the flat Box vs ElevatedCard decision in `TimelineCard`.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/TimelineItem.kt
joelmuraguri: No we should include for Tree and Linear threads in this case

`Threaded.Tree` is always the anchor so it belongs in `isThreadedAnchor` then `isThreadedAncestorOrAnchor` stays clean and derived

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/recordResolver/MutableTimelineItemCreationContext.kt
tunjid: Okay, let's do that then

tunjid: Actually, lets leave it for now. I really should not have leaked the `list` as a `MutableList` either. We'll come up with a better DSL for it later.

joelmuraguri: Sounds good, will leave it as is and revisit as part of the DSL refactor.

PR #1120, opened 2026-03-19
Gstreamer video player for linux desktop

joelmuraguri:
**add gst1-java-core dependency and linux javafx classifier**
- Add `gstreamer.java.core` to desktopMain dependencies
- Add 'linux' to the javafx classifier resolution so JavaFX JARs still compile on Linux even though they are not used at runtime
- Add gstreamer-java version to libs.versions.toml

**add GStreamer singleton initializer**
- `GStreamer.kt` wraps `Gst.init()` behind a double-checked lock so it is safe to call from any thread and only runs once per process

**implement `GStreamerPlayerState` and `GStreamerPlayerController` for Linux** 
`GStreamerPlayerState`
- Mirrors `AVFoundationPlayerState` double-buffered Skia bitmaps, serial dispatcher for frame writes, copyBgraFrame reused as-is
- AppSink pulls BGRA frames and publishes them as ImageBitmap via currentFrame - Position tracked by a coroutine polling at 250ms (no native time callback)
- Bus listeners handle STATE_CHANGED, EOS and ERROR
- dispose() / reinitialize() lifecycle mirrors AVFoundation

`GStreamerPlayerController`
- Mirrors `AVFoundationPlayerController` lazy pipeline creation on first play(), same pause/seek/retry/unregisterAll shape
- `isMuted` changes propagated via snapshotFlow

**add `GStreamerPlayerState` branch to `VideoPlaye`r composable**
- Add `GStreamerVideoPlayer` composable using `Canvas` + `scaleAndAlignTo`, same rendering path as `AVFoundationVideoPlayer`
- Add `canShowVideo` / `canShowStill` helpers for `GStreamerPlayerState`
- Add `GStreamerPlayerState` branch to the VideoPlayer when expression

**route Linux to GStreamerPlayerController in createAppState**
- Extend `videoPlayerController` lambda to check `os.name` and include `GStreamerPlayerController` for Linux

### DIFF
--- a/ui/media/src/desktopMain/kotlin/com/tunjid/heron/media/video/linux/GStreamerPlayerState.kt
+++ b/ui/media/src/desktopMain/kotlin/com/tunjid/heron/media/video/linux/GStreamerPlayerState.kt
@@ -98,6 +98,9 @@ internal class GStreamerPlayerState(
 
     private val serialDispatcher = Dispatchers.Default.limitedParallelism(1)
     private var positionPollingJob: Job? = null
+    private var stateChangedConnection: Bus.STATE_CHANGED? = null
+    private var eosConnection: Bus.EOS? = null
+    private var errorConnection: Bus.ERROR? = null
 
     internal fun createPipeline(url: String) {
         val sink = buildAppSink()
@@ -120,20 +123,27 @@ internal class GStreamerPlayerState(
     }
 
     private fun connectBus(pb: PlayBin) {
-        pb.bus.connect { source, old, current, _ ->
+        stateChangedConnection = Bus.STATE_CHANGED { source, old, current, _ ->
             if (source === pb) onStateChanged(pb, old, current)
-        }
-        pb.bus.connect(
-            Bus.EOS { _ ->
-                onEndOfStream(pb)
-            },
-        )
-        pb.bus.connect(
-            Bus.ERROR { _, _, _ ->
-                stopPositionPolling()
-                status = PlayerStatus.Idle.Initial
-            },
-        )
+        }.also { pb.bus.connect(it) }
+
+        eosConnection = Bus.EOS { _ ->
+            onEndOfStream(pb)
+        }.also { pb.bus.connect(it) }
+
+        errorConnection = Bus.ERROR { _, _, _ ->
+            stopPositionPolling()
+            status = PlayerStatus.Idle.Initial
+        }.also { pb.bus.connect(it) }
+    }
+
+    private fun disconnectBus(pb: PlayBin) {
+        stateChangedConnection?.let { pb.bus.disconnect(it) }
+        eosConnection?.let { pb.bus.disconnect(it) }
+        errorConnection?.let { pb.bus.disconnect(it) }
+        stateChangedConnection = null
+        eosConnection = null
+        errorConnection = null
     }
 
     private fun onStateChanged(pb: PlayBin, old: State, current: State) {
@@ -268,8 +278,11 @@ internal class GStreamerPlayerState(
         appSink = null
         isPipelineInitialized = false
 
-        pb?.stop()
-        pb?.dispose()
+        if (pb != null) {
+            disconnectBus(pb)
+            pb.stop()
+            pb.dispose()
+        }
 
         skiaBitmapA?.close()
         skiaBitmapB?.close()


composeApp/src/desktopMain/kotlin/com/tunjid/heron/Platform.jvm.kt
```diff
@@ -46,12 +47,17 @@ fun createAppState(): AppState =
...
+                System.getProperty("os.name").startsWith("Mac") -> AVFoundationPlayerController(
+                    appMainScope = appMainScope,
+                )
+                System.getProperty("os.name").startsWith("Linux") -> GStreamerPlayerController(
+                    appMainScope = appMainScope,
+                )
+                else -> JavaFxPlayerController(
+                    appMainScope = appMainScope,
+                )
+            }
```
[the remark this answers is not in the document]
tunjid: We should do this.

ui/media/src/desktopMain/kotlin/com/tunjid/heron/media/video/linux/GStreamer.kt
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Wraps GStreamer initialisation so it only runs once per process.
+ * [Gst.init] is not idempotent across all versions of gst1-java-core,
+ * so we guard it ourselves.
+ */
+internal object GStreamer {
+    @Volatile private var initialized = false
+
+    fun ensureInitialized() {
```
tunjid: If we made the `Gstreamer` a class, could we do:

I think this can be replaced by a lazy delegate:

```
internal object GStreamer {
    val initialized by lazy {
        Gst.init(Version.BASELINE, "KMPVideoPlayer")
        true
    }
}
```

Then you can check if it's initialized before using `snapshotFlow { isMuted }`

ui/media/src/desktopMain/kotlin/com/tunjid/heron/media/video/linux/GStreamerPlayerController.kt
```diff
@@ -0,0 +1,156 @@
...
+        stateToRetry.dispose()
+        stateToRetry.reinitialize()
+        stateToRetry.createPipeline(stateToRetry.videoUrl)
+        stateToRetry.playBin?.play()
+    }
+
+    override fun unregisterAll(retainedVideoIds: Set<String>): Set<String> {
+        val toDispose = idsToStates.filterNot { retainedVideoIds.contains(it.key) }
+        toDispose.keys.forEach { idsToStates.remove(it) }
+        toDispose.values.forEach { it.dispose() }
```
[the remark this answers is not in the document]
tunjid: This seems like a good suggestion

ui/media/src/desktopMain/kotlin/com/tunjid/heron/media/video/linux/GStreamerPlayerState.kt
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    private fun buildAppSink(): AppSink = AppSink("videosink").apply {
+        set("emit-signals", true)
+        caps = Caps.fromString("video/x-raw,format=BGRA")
+        connect(AppSink.NEW_SAMPLE { sink -> onNewSample(sink) })
+    }
+
+    private fun connectBus(pb: PlayBin) {
+        pb.bus.connect { source, old, current, _ ->
```
tunjid: Do these connection methods have ways to unregister the connectors?

joelmuraguri: Yes, `Bus.connect()` returns the listener which can be passed back to `Bus.disconnect()`. Added references for all three connections (`STATE_CHANGED`, `EOS`, `ERROR`) and a `disconnectBus()` that clears them. It's called at the top of `dispose()` before `pb.stop()`.

ui/media/src/desktopMain/kotlin/com/tunjid/heron/media/video/linux/GStreamer.kt
joelmuraguri: Done, replaced the double-checked lock with a `lazy` delegate and access `GStreamer.initialized` in the controller `init` block to trigger it.

joelmuraguri:
> There's a few things that we can work through later, but for now this works.

Sounds good, happy to go through them whenever you're ready.

PR #1158, opened 2026-03-28
Gstreamer player improvements

joelmuraguri:
GStreamer player fixes and Linux build task ordering

### 1. Disconnect AppSink.NEW_SAMPLE listener on dispose
The `NEW_SAMPLE` listener was connected anonymously with no reference
held, making it impossible to remove. Stored the connection on the
state and call `appSink.disconnect()` inside `dispose()`, matching the
same pattern used for the bus listeners.

### 2. Pause pipeline before seeking on fresh start
When resuming from a saved position on a fresh pipeline, play was
firing before seek. GStreamer requires the pipeline to be in `PAUSED`
state before a seek will land. Fixed by pausing first, seeking,
then calling play.

### 3. Fix JNA extraction task ordering on Linux
The `extractJnaNativeArm` and `extractJnaNativeX64` tasks were only
declared as dependencies of `desktopProcessResources` inside the Mac
condition. On Linux, Gradle 9 strict task ordering caught this as
an implicit dependency violation and failed the build. Moved the
JNA extraction dependsOn outside the Mac block so it runs on all
desktop platforms.

### 4. Add Linux JNA extraction tasks with platform onlyIf guards
This is similary tied to `3`
- Add extractJnaNativeLinuxX64 and extractJnaNativeLinuxArm tasks
  to extract libjnidispatch.so from linux-* paths in the JNA JAR
- Declare all four extraction tasks in desktopProcessResources
  unconditionally to satisfy Gradle 9 implicit dependency checks
- Add onlyIf guards on each task so Mac tasks skip on Linux and
  Linux tasks skip on Mac
- Keep original Mac task names so composeApp references are unchanged

### DIFF
--- a/ui/media/build.gradle.kts
+++ b/ui/media/build.gradle.kts
@@ -129,18 +129,17 @@ kotlin.sourceSets.named("desktopMain") {
 }
 
 tasks.named("desktopProcessResources") {
-    // Declare all extraction tasks so Gradle 9 implicit dependency
-    // checks pass onlyIf guards on each task handle actual execution
-    dependsOn(
-        "extractJnaNativeArm",
-        "extractJnaNativeX64",
-        "extractJnaNativeLinuxX64",
-        "extractJnaNativeLinuxArm",
-    )
-    if (System.getProperty("os.name").startsWith("Mac")) {
-        dependsOn(
+    val osName = System.getProperty("os.name")
+    when {
+        osName.startsWith("Mac") -> dependsOn(
             "buildAVFoundationMacArm",
             "buildAVFoundationMacX64",
+            "extractJnaNativeArm",
+            "extractJnaNativeX64",
+        )
+        osName.startsWith("Linux") -> dependsOn(
+            "extractJnaNativeLinuxX64",
+            "extractJnaNativeLinuxArm",
         )
     }
 }


ui/media/build.gradle.kts
```diff
@@ -143,12 +143,16 @@ kotlin.sourceSets.named("desktopMain") {
 }
 
 tasks.named("desktopProcessResources") {
+    // JNA native extraction needed on all desktop platforms
+    dependsOn(
+        "extractJnaNativeArm",
+        "extractJnaNativeX64",
+    )
```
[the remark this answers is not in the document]
joelmuraguri: Good catch. Fixed by adding platform-specific Linux extraction tasks (extractJnaNativeLinuxX64, extractJnaNativeLinuxArm) that pull libjnidispatch.so from linux-* paths. All four tasks are now declared in desktopProcessResources unconditionally to satisfy Gradle 9, but each has an onlyIf guard so Mac tasks skip on Linux and vice versa. Mac task names are unchanged so composeApp references are unaffected.

ui/media/build.gradle.kts
```diff
@@ -144,12 +142,18 @@ kotlin.sourceSets.named("desktopMain") {
 }
 
 tasks.named("desktopProcessResources") {
+    // Declare all extraction tasks so Gradle 9 implicit dependency
+    // checks pass onlyIf guards on each task handle actual execution
+    dependsOn(
```
tunjid: Should we be checking what os this is?

joelmuraguri: Yes this was necessary because `./gradlew :composeApp:run` failed without it.

joelmuraguri: Is there a better alternative? I think what we had initially supported for mac desktop only

tunjid: We could only run `extractJnaNativeLinuxX64` and `extractJnaNativeLinuxArm` if the system building the app is a linux machine and we're building the linux app on CI. ATM those tasks will run regardless of the machine building it

joelmuraguri: Done and works. Agree with the approach, Linux tasks now only declared as dependencies when building on Linux, same pattern as Mac.

tunjid:
@joelmuraguri consider writing a gradle convention plugin that reports the current OS of the build machine so it can be applied in the `ui:media` and `composeApp` modules.

PR #1192, opened 2026-04-09
FeedList membership writable actions

joelmuraguri:
- Adds `FeedList.UpdateProfileList` sealed class with `Add` and `Remove` actions
- Implements `updateProfileList` in `ProfileRepository` handling both actions via AT Protocol
- Wires `Writable.UpdateProfileList` following the existing polymorphic pattern

UI layer to follow

### DIFF
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
@@ -29,6 +29,7 @@ import app.bsky.graph.GetFollowsQueryParams
 import app.bsky.graph.GetFollowsResponse
 import app.bsky.graph.GetMutesQueryParams
 import app.bsky.graph.GetMutesResponse
+import app.bsky.graph.Listitem
 import app.bsky.graph.MuteActorRequest
 import app.bsky.graph.UnmuteActorRequest
 import com.atproto.repo.CreateRecordRequest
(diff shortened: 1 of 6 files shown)


data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
```diff
@@ -544,4 +555,55 @@ internal class OfflineProfileRepository @Inject constructor(
...
+                            uri = it.uri.atUri.let(::ListMemberUri),
+                            listUri = update.listUri,
+                            subjectId = update.profileId,
+                            createdAt = Clock.System.now(),
+                        ),
+                    ),
+                )
+            }
+
+            is FeedList.UpdateProfileList.Remove -> networkService.runCatchingWithMonitoredNetworkRetry {
```
tunjid: For removing a profile from a List, we should use `RecordRepository.deleteRecord`. It has central logic for deleting records. We should keep consolidating on it. The path for adding a member is right.

tunjid: You can check for usages of deleteRecord to see what I mean.

joelmuraguri: Thanks for pointing this out!

I have made updates;
- Defined a data class `AddProfileList` for handling add a member writable action
- Implemented `addProfileList` in `ProfileRepository` handles for adding a member
- Wire `Writable.AddProfileList` in write queue

tunjid: Ah looks like a class already exists for creating a list member too? https://github.com/tunjid/heron/blob/main/data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/models/ListMember.kt#L40-L44

It also seems like this is wired up. I think your changes just need to be UI side.

PR #1202, opened 2026-04-12
Add/Remove Profile from Lists

joelmuraguri:
## What
Allows a signed-in user to add or remove any profile from their own lists directly from the profile screen, via a bottom sheet picker.

### UI
- Added a new entry in `ProfileActionsMenu` to trigger the list membership editor
- Implemented `ListMemberPickerBottomSheet` shows the signed-in user's lists with a `FilterChip` per item indicating current membership status (Add/Remove)
- `ListMemberPickerItem` uses `RecordLayout` with `ListMembershipChip` driven by a nullable `ListMember?`  presence means member, null means not

### Data
- Added `membershipsByProfile(profileId)` DAO query on `listMembers` table scoped to a subject profile
- Implemented `membershipsByProfile` in the repository using `singleAuthorizedSessionFlow`, returning `Flow<List<ListMember>>` for the viewed profile

### ViewModel / State
- Added `profileListMemberships: List<ListMember>` and `signedInProfileListsHolder` to `State`
- `loadProfileMutations` now uses `flatMapLatest` to resolve `profile.did` before fetching memberships avoids unsafe `ProfileHandleOrId` cast
- Added `Action.AddListMember` with a corresponding `enqueueMutation` writing `FeedList.AddMember` to the `WriteQueue`
- Removal reuses the existing `Action.DeleteRecord` since `ListMemberUri` is a `RecordUri`

## Why
The lists data was already loaded per profile tab via `ProfileScreenStateHolders.Records.Lists`  the signed-in user's lists are sourced from a dedicated `signedInProfileListsHolder` so the correct lists always appear regardless of which profile is being viewed.

## Demo

https://github.com/user-attachments/assets/ce2cd9fa-4e8e-40b2-8357-8186e362c1f3

### DIFF
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
@@ -40,7 +40,6 @@ import com.tunjid.heron.data.core.models.Cursor
 import com.tunjid.heron.data.core.models.CursorList
 import com.tunjid.heron.data.core.models.CursorQuery
 import com.tunjid.heron.data.core.models.DataQuery
-import com.tunjid.heron.data.core.models.ListMember
 import com.tunjid.heron.data.core.models.Profile
 import com.tunjid.heron.data.core.models.ProfileViewerState
 import com.tunjid.heron.data.core.models.ProfileWithViewerState
@@ -48,7 +47,6 @@ import com.tunjid.heron.data.core.models.value
 import com.tunjid.heron.data.core.types.BlockUri
 import com.tunjid.heron.data.core.types.FollowUri
 import com.tunjid.heron.data.core.types.Id
-import com.tunjid.heron.data.core.types.ProfileId
 import com.tunjid.heron.data.core.types.RecordCreationException
 import com.tunjid.heron.data.core.types.Uri
 import com.tunjid.heron.data.core.types.recordKey
@@ -114,10 +112,6 @@ interface ProfileRepository {
         limit: Long,
     ): Flow<List<Profile>>
 
-    fun membershipsByProfile(
-        profileId: ProfileId,
-    ): Flow<List<ListMember>>
-
     fun followers(
         query: ProfilesQuery,
         cursor: Cursor,
@@ -153,7 +147,6 @@ interface ProfileRepository {
 internal class OfflineProfileRepository @Inject constructor(
     @param:IODispatcher
     private val ioDispatcher: CoroutineDispatcher,
-    private val listDao: ListDao,
     private val profileDao: ProfileDao,
     private val profileLookup: ProfileLookup,
     private val networkService: NetworkService,
@@ -219,22 +212,6 @@ internal class OfflineProfileRepository @Inject constructor(
             )
                 .distinctUntilChangedMap { profileEntities ->
                     profileEntities.map(PopulatedProfileEntity::asExternalModel)
-                }
-        }
-            .flowOn(ioDispatcher)
-
-    override fun membershipsByProfile(
-        profileId: ProfileId,
-    ): Flow<List<ListMember>> =
-        savedStateDataSource.singleAuthorizedSessionFlow { signedInProfileId ->
-            if (profileId.id.isEmpty()) return@singleAuthorizedSessionFlow emptyFlow<List<ListMember>>()
-
-            listDao.membershipsByProfile(
-                profileId = profileId.id,
-                signedInUserId = signedInProfileId.id,
-            )
-                .distinctUntilChangedMap { entities ->
-                    entities.map { it.asExternalModel() }
                 }
         }
             .flowOn(ioDispatcher)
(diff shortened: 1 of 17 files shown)


feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ui/ListMemberPickerSheetState.kt
```diff
@@ -0,0 +1,136 @@
...
+class ListMemberPickerSheetState(
+    scope: BottomSheetScope,
+) : BottomSheetState(scope) {
+
+    override fun onHidden() {}
+
+    companion object {
+        @Composable
+        fun rememberListMemberPickerSheetState(
+            paneTransitionScope: PaneTransitionScope,
```
tunjid: Compose modal bottom sheets do not share the same window with the app, they are dialogs backed by a separate window and therefor a separate window token. Due to this dialogs cannot have shared elements with the rest of the app, making the `paneTransitionScope` have no effect.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/list/FeedList.kt
```diff
@@ -97,6 +111,55 @@ fun FeedList(
     )
 }
 
+@Composable
+fun ListMemberPickerItem(
```
tunjid: List members are profile oriented, so the `ProfileWithViewerState` Composable in the `com.tunjid.heron.timeline.ui.profile` package is more appropriate. If you go to the `ListsScreen`, the people tab has a layout for displaying list members. I think we should consolidate around this template.

tunjid: Oh I see, this is for displaying lists that we may want edit memberships of.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/ProfileRepository.kt
```diff
@@ -215,6 +223,22 @@ internal class OfflineProfileRepository @Inject constructor(
         }
             .flowOn(ioDispatcher)
 
+    override fun membershipsByProfile(
+        profileId: ProfileId,
+    ): Flow<List<ListMember>> =
+        savedStateDataSource.singleAuthorizedSessionFlow { signedInProfileId ->
```
tunjid: This just fetches from the db with no network refresh. If you open the bluesky website and inspect network traffic, when adding a profile to the lists of the signed in user, they make a call to the API to the signed in profiles list members records to fetch all the list members the signed in user has created.

This is the only way they are able to tell if a profile already exists in a list or not. Its inefficient, but what we have to do to.

There should also be a createdListMembers method, and should be exposed on `RecordRepository` via `BlueskyRecordOperations`. It should list the `listRecords` XRPC to paginate list members on the `app.bsky.graph.listitem` collection. Only then can we verify that a profile is a member of a list or not.

joelmuraguri: Okay I hear you and your approach makes sense, I have made changes regarding this.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/records/BlueskyRecordOperations.kt
```diff
@@ -357,6 +364,50 @@ internal class OfflineFirstBlueskyRecordOperations @Inject constructor(
...
             .flowOn(ioDispatcher)
 
+    override fun createdListMembers(
+        query: CreatedListMembersQuery,
+        cursor: Cursor,
+    ): Flow<CursorList<ListMember>> =
+        savedStateDataSource.singleAuthorizedSessionFlow { signedInProfileId ->
+            combine(
+                listDao.listMembersByProfile(
+                    profileId = query.profileId.id,
```
tunjid: The pagination here will likely break as this db query does not match the network result. The DB query should fetch list members created by the signed in profile, it shouldn't need a profile Id.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/RecordRepository.kt
```diff
@@ -75,6 +93,37 @@ internal class OfflineFirstRecordRepository @Inject constructor(
...
+                        query = CreatedListMembersQuery(
+                            profileId = signedInProfileId,
+                            data = CursorQuery.Data(
+                                page = 0,
+                                cursorAnchor = Clock.System.now(),
+                                limit = 15,
+                            ),
+                        ),
+                        cursor = Cursor.Initial,
+                    ).first()
```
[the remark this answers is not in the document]
tunjid: This is valid.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/multipleEntitysaver/SaveListItemView.kt
```diff
@@ -39,3 +40,17 @@ internal fun MultipleEntitySaver.add(
         ),
     )
 }
+
+internal fun MultipleEntitySaver.add(
+    listMemberUri: ListMemberUri,
+    listItem: Listitem,
+) {
+    add(
+        ListMemberEntity(
```
tunjid, line 50: Trying to save a list member without saving the profiles of the list members will cause errors because of failed foreign key constraints.

tunjid (commented):
Think a little big picture about this. You want to add or remove a profile from the signed in profiles lists. To do this reliably, you want to present to the signed in user if a profile is in their lists.

* What are the things you need to know to present this?
* Do you have a reliable way of getting this information?

PR #1242, opened 2026-04-22
Show and apply provider handle suffix in sign-in

### LINKED ISSUE #1220: Auto-append provider domain for username-only login input
In provider-based login flows, users may enter only the username part and not realize they need to enter the full handle including the domain. This can create unnecessary friction during onboarding.

If a user enters only the username part, automatically append the selected provider’s default handle domain before validation and resolution.

For example, in the Bluesky flow:
- `alice` -> `alice.bsky.social`

If the user already enters a full handle including the domain, it should be preserved as-is.

- Requiring users to always enter the full handle including the domain

**Additional context**
This would make the login flow more intuitive, especially for new users who are more likely to think in terms of a username than a handle with a domain.

HI-JIN2:
## Summary
- Resolved #1220 
- append the selected provider's default handle suffix when users enter only a username
- preserve fully qualified handles as-is
- show the inferred suffix inline in the sign-in field for both OAuth and password flows

[Screen_recording_20260422_212746.webm](https://github.com/user-attachments/assets/44c0799e-2df6-4785-ad2b-6b7857c67a1e)

### DIFF
--- a/feature/auth/src/commonMain/kotlin/com/tunjid/heron/signin/SignInScreen.kt
+++ b/feature/auth/src/commonMain/kotlin/com/tunjid/heron/signin/SignInScreen.kt
@@ -126,8 +126,7 @@ internal fun SignInScreen(
                                     )
                                 }
                             }
-                        }
-                        else null,
+                        } else null,
                         onValueChange = { field, newValue ->
                             actions(
                                 Action.FieldChanged(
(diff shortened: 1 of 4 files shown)


tunjid:
Thanks so much for this! I'm still reviewing, sorry its a bit slow. I'm taking my time bc it affects all users of the app when signing in. Any bugs could mean sign out is broken

PR #1267, opened 2026-05-02
Embed Record links when composing a post 

joelmuraguri:
Pasting a bsky.app link while composing a post or reply now automatically embeds the linked record matching the behaviour in the official Bluesky app.

What changed
- `bskyHttpsUrlToAtUri()` converts any `https://bsky.app/`. URL to its `at://` equivalent, covering posts, feeds, lists, starter packs and labelers
- `asEmbeddableRecordUriOrNull()`  now handles both `https://` and `at://` inputs, so the rest of the pipeline needs no special casing
- `PostComposition` scans all parsed links for any ExternalLink and surfaces it via `onExternalLinkDetected`
- `Action.EmbedUrl` + `embedUrlMutations` resolves the URL through `recordRepository.embeddableRecord()` and updates `state.embeddedRecord`
- `ComposeScreen` wires `onExternalLinkDetected` to dispatch `Action.EmbedUrl`, guarded against overwriting an already embedded record

## Demo

https://github.com/user-attachments/assets/e0045e23-da57-4af5-a6fb-db1439a398e8

### DIFF
--- a/data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/types/Uri.kt
+++ b/data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/types/Uri.kt
@@ -365,10 +365,14 @@ fun String.asRecordUriOrNull(): RecordUri? = atUriComponents { _, collectionRang
     }
 }
 
-fun String.asEmbeddableRecordUriOrNull(): EmbeddableRecordUri? =
-    atUriComponents { _, collectionRange, _ ->
-        val normalized = withAtProtoPrefix()
-        when (substring(collectionRange.start, collectionRange.endExclusive)) {
+fun String.asEmbeddableRecordUriOrNull(): EmbeddableRecordUri? {
+    val atUri = when {
+        startsWith(Uri.Host.Https.prefix) -> bskyHttpsUrlToAtUri() ?: return null
+        else -> this
+    }
+    return atUri.atUriComponents { _, collectionRange, _ ->
+        val normalized = atUri.withAtProtoPrefix()
+        when (atUri.substring(collectionRange.start, collectionRange.endExclusive)) {
             PostUri.NAMESPACE -> PostUri(normalized)
             FeedGeneratorUri.NAMESPACE -> FeedGeneratorUri(normalized)
             ListUri.NAMESPACE -> ListUri(normalized)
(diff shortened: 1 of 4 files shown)


data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/types/Uri.kt
```diff
@@ -377,11 +381,43 @@ fun String.asEmbeddableRecordUriOrNull(): EmbeddableRecordUri? =
...
+    "labeler" to LabelerUri.NAMESPACE,
+)
+
+fun String.bskyHttpsUrlToAtUri(): String? {
+    if (!startsWith(Uri.Host.Https.prefix)) return null
+    val segments = removePrefix(Uri.Host.Https.prefix)
+        .removePrefix("bsky.app/")
+        .split("/")
+
+    val (handle, type, rkey) = when (segments.size) {
```
tunjid: Can we do this without allocating? atUri.atUriComponents doesn't allocate. We've had performance issues for the timeline in the past, this one is relatively minor but if we can avoid allocating, we should.

feature/compose/src/commonMain/kotlin/com/tunjid/heron/compose/ComposeViewModel.kt
```diff
@@ -169,6 +173,19 @@ private fun embeddedRecordMutations(
...
+): Flow<Mutation<State>> =
+    mapLatestToManyMutations { action ->
+        val uri = action.url.asEmbeddableRecordUriOrNull() ?: return@mapLatestToManyMutations
+        emitAll(
+            embeddedRecordMutations(
+                embeddedRecordUri = uri,
+                recordRepository = recordRepository,
+            ),
+        )
+    }
```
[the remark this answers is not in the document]
tunjid: We should probably limit emissions with a `take(1)` or `first()`. We don't want to keep emitting the mutation for every change to the underlying record.

data/models/src/commonMain/kotlin/com/tunjid/heron/data/core/types/Uri.kt
joelmuraguri: It's done. I've refactored to use index-based substring lookups mirroring atUriComponents, no `split()`, `filter()`, or intermediate list allocations.

PR #1282, opened 2026-05-06
RockSky Local DB Entities & AutoMigration

joelmuraguri:
Add local Room database entities for RockSky.

**New entities:**
- `RockSkyArtistEntity`
- `RockSkyAlbumEntity`
- `RockSkyTrackEntity`
- `RockSkyScrobbleEntity`

**Migration:** Database version bumped 40 to 41 via `@AutoMigration(from = 40, to = 41)`.


data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/RockSkyAlbumEntity.kt
```diff
@@ -0,0 +1,75 @@
...
+import androidx.room.PrimaryKey
+import androidx.room.Relation
+import com.tunjid.heron.data.core.models.RockSkyAlbum
+import com.tunjid.heron.data.core.types.AlbumId
+import com.tunjid.heron.data.core.types.AlbumUri
+import com.tunjid.heron.data.core.types.ArtistUri
+import com.tunjid.heron.data.core.types.ImageUri
+
+@Entity(
+    tableName = "rockSkyAlbums",
```
tunjid: For these, lets stylize "rocksky" as all lowercase, so `rockskyAlbums`

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/RockSkyAlbumEntity.kt
```diff
@@ -0,0 +1,75 @@
...
+            onDelete = ForeignKey.CASCADE,
+            onUpdate = ForeignKey.CASCADE,
+        ),
+    ],
+    indices = [
+        Index(value = ["uri"]),
+        Index(value = ["artistUri"]),
+    ],
+)
+data class RockSkyAlbumEntity(
```
tunjid: Same for these, should be `RockskyAlbumEntity`

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/RockSkyScrobbleEntity.kt
```diff
@@ -0,0 +1,95 @@
...
+    indices = [
+        Index(value = ["uri"]),
+        Index(value = ["did"]),
+        Index(value = ["trackUri"]),
+        Index(value = ["artistUri"]),
+        Index(value = ["albumUri"]),
+        Index(value = ["createdAt"]),
+    ],
+)
+data class RockSkyScrobbleEntity(
```
tunjid: For these, we need composite primary keys. We need to add the profile did of the profile that listened. The app supports multiple users, we needs scrobbles to be unique per user.

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/RockSkyScrobbleEntity.kt
```diff
@@ -0,0 +1,95 @@
...
+    val uri: ScrobbleUri,
+    val cid: ScrobbleId,
+    val trackId: TrackId,
+    val title: String,
+    val artist: String,
+    val albumArtist: String?,
+    val album: String?,
+    val albumArt: ImageUri?,
+    val handle: ProfileHandle?,
+    val did: ProfileId?,
```
tunjid: I think the `did` has to be required.

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/RockSkyAlbumEntity.kt
```diff
@@ -0,0 +1,75 @@
...
+    @PrimaryKey
+    val uri: AlbumUri,
+    val cid: AlbumId,
+    val title: String,
+    val artist: String,
+    val releaseDate: String?,
+    val year: Int?,
+    val albumArt: ImageUri?,
+    val artistUri: ArtistUri?,
```
tunjid: Should this not be required?

joelmuraguri: For some albums The API itself doesn't guarantee it but every album must have an artist in the real sense. I think keeping it nullable on the entity to match what the API can return is the right call. Do you we make it required?

joelmuraguri: I have made it required, although for some albums, the API doesnt guarantee.

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/RockskyScrobbleEntity.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.tunjid.heron.data.core.types.ProfileId
+import com.tunjid.heron.data.core.types.ScrobbleId
+import com.tunjid.heron.data.core.types.ScrobbleUri
+import com.tunjid.heron.data.core.types.TrackId
+import com.tunjid.heron.data.core.types.TrackUri
+import kotlin.time.Instant
+
+@Entity(
+    tableName = "rockskyScrobbles",
+    primaryKeys = ["uri", "viewerDid"],
```
tunjid: Oh I didn't mean the viewerId here, I meant the did; i.e the user who scrobbled. However I do thaink I made a mistake as it will already be unique on the did as the uri will have the scrobblers did. Sorry about that

tunjid (approved):

PR #1297, opened 2026-05-12
Rocksky Offlline-first persistence

joelmuraguri:
Wires up full offline-first persistence for Rocksky following the
existing combine(localFlow, networkCursorFlow) pattern.

- Add RockskyDao with upserts, paginated reads and deletes
- Add entities in `MultipleEntitySaver` and `MultipleEntitySaverProvider`
- Map network views to entities via `MultipleEntitySaver` extensions
- Replace network-only `paged()` helper with offline-first combine pattern
- Scope all data per user via creatorId
- AutoMigration 42-43 (new tables), manual migration 41-42 (creatorId column)
- Rename RockSky - Rocksky across all class names

 Rocksky API is currently unavailable so network responses will be
empty in the interim, offline reads from Room will work once data
has been seeded.


data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/entities/RockskyAlbumEntity.kt
```diff
@@ -26,12 +27,14 @@ import com.tunjid.heron.data.core.types.ImageUri
...
         Index(value = ["uri"]),
         Index(value = ["artistUri"]),
+        Index(value = ["creatorId"]),
     ],
 )
 data class RockskyAlbumEntity(
     @PrimaryKey
     val uri: AlbumUri,
     val cid: AlbumId,
+    val creatorId: ProfileId,
```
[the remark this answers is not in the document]
joelmuraguri: This app persists a single user session at a time, so multiple local users interacting with the same album URI is not a scenario we need to handle. I think uri as sole PK is correct here.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/multipleEntitysaver/SaveRockskyAlbumView.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.tunjid.heron.data.core.types.ArtistUri
+import com.tunjid.heron.data.core.types.ImageUri
+import com.tunjid.heron.data.core.types.ProfileId
+import com.tunjid.heron.data.database.entities.RockskyAlbumEntity
+
+internal fun MultipleEntitySaver.add(
+    creatorId: ProfileId,
+    albumView: AlbumView,
+) {
+    add(
```
tunjid, line 17: We need to also add a stub of the creator in case its not in the db

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/multipleEntitysaver/SaveRockskyArtistView.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.tunjid.heron.data.core.types.ArtistUri
+import com.tunjid.heron.data.core.types.ImageUri
+import com.tunjid.heron.data.core.types.ProfileId
+import com.tunjid.heron.data.database.entities.RockskyArtistEntity
+
+internal fun MultipleEntitySaver.add(
+    creatorId: ProfileId,
+    artistView: ArtistView,
+) {
+    add(
```
tunjid, line 16: Same about stubbing the creator.

tunjid:
@joelmuraguri sorry about the delay in looking at this, I'm chasing down a pretty bad memory leak. The Rocky API should be a bit more stable now so you should be able to test the network. Visit this profile to verify records: https://bsky.app/profile/tsiry-sandratraina.com

joelmuraguri:
> @joelmuraguri sorry about the delay in looking at this, I'm chasing down a pretty bad memory leak. The Rocky API should be a bit more stable now so you should be able to test the network. Visit this profile to verify records: https://bsky.app/profile/tsiry-sandratraina.com

Thanks for the heads up! Verified with the profile and the API is working great, data is coming through nicely.

data/database/src/commonMain/kotlin/com/tunjid/heron/data/database/migrations/Migration42To43RockskyCreatorId.kt
```diff
@@ -0,0 +1,31 @@
+package com.tunjid.heron.data.database.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.SQLiteConnection
+import androidx.sqlite.execSQL
+
+internal object Migration42To43RockskyCreatorId : Migration(42, 43) {
```
tunjid, line 7: This works, and is correct, however `AutoMigration` can add columns.

tunjid (approved):

PR #1321, opened 2026-05-19
Auto focus search bar on screen entry and nav selection

### LINKED ISSUE #1314: Focus the search bar on the search screen when the bottom nav is reselected
See https://github.com/tunjid/heron/issues/1311

The PaneNavigationBar has a `[onNavItemReselectedCallback`](https://github.com/tunjid/heron/blob/93543b59396465ccb240220d7034562092354341/scaffold/src/commonMain/kotlin/com/tunjid/heron/scaffold/scaffold/PaneNavigation.kt#L76). Tapping it should focus the search bar. The `SearchBar` composable currently has a `FocusRequester` local to it. You will most likely need to hoist it, so `onNavItemReselectedCallback` can be used to request focus.

joelmuraguri:
Resolves #1314. Users navigating to the Search screen or re-clicking
the Search navigation item expect the search input field to be immediately focused with the keyboard open.

## Changes

- Hoisted `FocusRequester` within `SearchBindings`.
- Added a `LaunchedEffect` to automatically request focus on the initial screen composition.
- Bound the hoisted `FocusRequester` to `PaneNavigationBar` 
 `(onNavItemReselected)` to gracefully restore focus whenever the search nav icon is tapped while already on the screen

### DIFF
--- a/feature/search/src/commonMain/kotlin/com/tunjid/heron/search/di/Bindings.kt
+++ b/feature/search/src/commonMain/kotlin/com/tunjid/heron/search/di/Bindings.kt
@@ -21,7 +21,10 @@ import androidx.compose.foundation.layout.offset
 import androidx.compose.material3.ExperimentalMaterial3ExpressiveApi
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
 import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.unit.round
 import androidx.lifecycle.viewmodel.compose.viewModel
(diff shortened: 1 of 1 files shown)


tunjid:
Let's just focus when the tab is reelected. Auto focusing would mean users won't see tge discover items most of the time as the keyboard will cover it. The official app also does it this way.

feature/search/src/commonMain/kotlin/com/tunjid/heron/search/di/Bindings.kt
```diff
@@ -143,6 +146,12 @@ class SearchBindings(
             val state = stateHolder.produceStateWithLifecycle()
             val paneScaffoldState = rememberPaneScaffoldState()
 
+            val searchFocusRequester = remember { FocusRequester() }
+
```
tunjid, line 150: Wil this request focus as soon as we enter the screen?

joelmuraguri: It request focus only when tab is reselected.

tunjid (approved):

PR #1343, opened 2026-06-01
Migrate MutedWordsSheetState to ViewModel-backed state

joelmuraguri:
Introduces per-sheet ViewModels for bottom sheets that require data  access, starting with MutedWordsSheetState as the reference implementation. Establishes the DI wiring pattern for all subsequent sheet migrations.

### How

**ViewModel**
- `MutedWordsViewModel` is @AssistedInject with `UserDataRepository` and WriteQueue provided by Metro, and only `CoroutineScope` as `@Assisted`.
- `MutedWordsViewModelInitialize` (`@AssistedFactory`) receives scope at runtime. State is a `@Serializable` data class backed by `StateFlow` via `actionStateFlowMutator`. 
- `Preferences` are seeded once via `take(1)` to prevent the live preferences flow from overwriting local edits.
Writes back via `WriteQueue` when the sheet closes.

**SheetState**
- `MutedWordsSheetState` is now mechanics-only (show/hide/onHidden).
`rememberUpdatedMutedWordsSheetState` takes zero parameters.

**DI**
- `SheetsViewModelInitializers` aggregates sheet ViewModel factories.
- `SheetBindings` (`@Includes DataBindings`) provides `SheetsViewModelInitializers` to the graph. 
- `AppState` receives `SheetsViewModelInitializers` via Metro.
- `PaneScaffoldState` extensions  will expose sheet creation to feature screens with zero parameter passing.


tunjid (commented):
This is a fantastic PR! I'll suggest some change in two categories:

1. Update in this PR: Changes that should be updated in this PR.
2. Update in follow up: Changes to land in a follow up PR, but before any other sheet migrations.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/mutedwords/MutedWordsStateHolder.kt
```diff
@@ -0,0 +1,187 @@
...
+                    mutedWordPreferences = action.mutedWordPreference,
+                ),
+            )
+        },
+    ) { _, memo ->
+        if (memo != null) emit { copy(messages = messages + memo) }
+    }
+
+@Serializable
+data class MutedWordsState(
```
tunjid, line 155: Fix in this PR: Lets use Snapshottable for the State. I didn't realize this is what you were referring to in your design doc.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/mutedwords/MutedWordsStateHolder.kt
```diff
@@ -0,0 +1,187 @@
...
+
+@AssistedInject
+class MutedWordsViewModel(
+    userDataRepository: UserDataRepository,
+    writeQueue: WriteQueue,
+    @Assisted scope: CoroutineScope,
+) : ViewModel(viewModelScope = scope),
+    MutedWordsStateHolder by scope.actionStateFlowMutator(
+        initialState = MutedWordsState(),
+        started = SharingStarted.Lazily,
```
tunjid: Fix in this PR: Lets use 

```
started = SharingStarted.WhileSubscribed(SheetWhileSubscribed)
```

Where `SheetWhileSubscribed` is defined in this module to be 3 seconds. While the ViewModel store is destroyed while out of composition, I;d much rather we capped the state production.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/mutedwords/MutedWordSheetState.kt
```diff
@@ -90,376 +99,341 @@ class MutedWordsSheetState(
...
-        mutedWords = mutedWords.filterNot { it.value == value }
-        error = null
+@Composable
+private fun MutedWordsScreen(
+    sheetState: MutedWordsSheetState,
+    initializer: MutedWordsViewModelInitializer,
+) {
+    val stateHolder: MutedWordsStateHolder = viewModel<MutedWordsViewModel> {
+        initializer.invoke(
+            scope = CoroutineScope(
```
tunjid: Fix in update PR: We should move `com.tunjid.heron.scaffold.scaffold.viewModelCoroutineScope` from the scaffold module, to the `ui:core` module. This will let us reuse `viewModelCoroutineScope()` here instead of recreating it like this here.

It will cause a noisy diff, hence why t should be a follow up.

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
```diff
@@ -1360,13 +1360,8 @@ private fun ProfileTimeline(
             actions(Action.SendPostInteraction(it))
         },
     )
-    val mutedWordsSheetState = rememberUpdatedMutedWordsSheetState(
```
tunjid, line 1363: Update in follow up PR: Being able to call through to the data layer from sheets genuinely makes me very happy. We should also delete `Action.UpdateMutedWord` from the profile screen's actions since it is no longer being used. Same for all the other screens that have been migrated.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/mutedwords/MutedWordsStateHolder.kt
```diff
@@ -0,0 +1,187 @@
...
+
+@AssistedFactory
+fun interface MutedWordsViewModelInitializer {
+    fun invoke(
+        scope: CoroutineScope,
+    ): MutedWordsViewModel
+}
+
+@AssistedInject
+class MutedWordsViewModel(
```
tunjid, line 42: Fix in this PR: Make this internal, metro should still be able to inject. Other modules should not be able to see this type.

tunjid: Actually never mind, it does need to be public

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/mutedwords/MutedWordSheetState.kt
```diff
@@ -90,376 +99,341 @@ class MutedWordsSheetState(
...
+        provider = storeProvider,
+        key = "muted_words_sheet",
+    )
 
-        // Reset input fields
-        newWord = ""
-        newWordDuration = null
-        error = null
+    state.ModalBottomSheet {
+        CompositionLocalProvider(LocalViewModelStoreOwner provides storeOwner) {
```
tunjid, line 132: This PR: I think you can move this setup inside `state.ModalBottomSheet`. I don't think we need to specify a key yet.

This way, any call to `state.ModalBottomSheet { }` will already have its own view model store provider set up for that point in the composition

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/mutedwords/MutedWordSheetState.kt
```diff
@@ -90,376 +99,341 @@ class MutedWordsSheetState(
...
+            )
+        }
     }
+}
 
-    fun removeMutedWord(value: String) {
-        mutedWords = mutedWords.filterNot { it.value == value }
-        error = null
+@Composable
+private fun MutedWordsScreen(
```
tunjid: This PR: Should this be `MutedWordsSheet`?

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/mutedwords/MutedWordsStateHolder.kt
tunjid: Oh I see you have this as a follow up already: https://github.com/tunjid/heron/issues/1344

Follow up is fine.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/mutedwords/MutedWordSheetState.kt
joelmuraguri: I've updated to `MutedWordsSheet`

feature/profile/src/commonMain/kotlin/com/tunjid/heron/profile/ProfileScreen.kt
joelmuraguri: I have made this changes in this PR.

tunjid (approved):

PR #1352, opened 2026-06-03
Consolidate AppState stateHolders for uniformity

tunjid:


ui/core/src/commonMain/kotlin/com/tunjid/heron/ui/coroutines/Utilities.kt
```diff
@@ -27,3 +32,25 @@ fun viewModelCoroutineScope() = CoroutineScope(
...
+        }
+    }
+
+    val snapshotObserver = Snapshot.registerGlobalWriteObserver {
+        snapshotChanged.trySend(Unit)
+    }
+
+    try {
+        block()
+    } finally {
```
ardakazanci, line 52: Why didn’t you choose to catch the error using `catch`? That way, it would be easy to spot any side effects that might occur.

tunjid: This method doesn't have anything that might throw. If `block` throws, its the responsibility of the caller to catch it. Catching it here would just swallow the error.

PR #1371, opened 2026-06-07
Migrate ThreadGateSheetState to ViewModel-backed state

joelmuraguri:
### What
Migrates `ThreadGateSheetState` to the ViewModel-backed pattern  Covers both `OfTimeline` and `OfPreference` subtypes.

### Why
`recentLists` was passed through multiple composable layers and
`onRequestRecentLists` coupled sheet UI interactions back to the
feature ViewModel. The ViewModel now loads lists directly from
`RecordRepository`, removing both.

### Write path why callbacks instead of ViewModel writes

Example in `MutedWordsViewModel` handles its own write because the operation is  always the same regardless of which screen opens the sheet. The ViewModel owns everything it needs.

`ThreadGateSheetState` is different it is opened from two contexts
with two completely different write operations:

- Timeline  `Action.SendPostInteraction`
- Compose/Moderation screen  `Action.UpdateInteractionSettings`

The sheet ViewModel has no way to know which context it is in. The
callback keeps write coordination at the feature level where it
belongs, and preserves the feature ViewModel's ability to handle
post-write feedback such as snackbar messages.

### Changes
- `ThreadGateViewModel` with `@AssistedInject`, `@Snapshottable` state, loads `recentLists` via `launchAndCollect` in the producer block
- `ThreadGateSheetState` receives the ViewModel directly; `mode` and`allowed` move into ViewModel state; `show()` dispatches
  `ThreadGateAction.Initialize`; `onHidden()` dispatches
  `ThreadGateAction.Reset`
- Two `PaneScaffoldState` extensions: `rememberTimelineThreadGateSheetState`
  and `rememberPreferenceThreadGateSheetState`
- `ComposeThreadGate` made a `PaneScaffoldState` extension; removes`recentLists` and `onRequestRecentLists`


feature/compose/src/commonMain/kotlin/com/tunjid/heron/compose/ui/ComposeThreadGate.kt
```diff
@@ -24,36 +24,28 @@ import androidx.compose.ui.Modifier
...
-        recentLists = recentLists,
-        onRequestRecentLists = onRequestRecentLists,
+    val threadGateSheetState = rememberPreferenceThreadGateSheetState(
         onDefaultThreadGateUpdated = onInteractionSettingsUpdated,
     )
 
     ElevatedCard(
         modifier = modifier,
         shape = CircleShape,
-        onClick = {
```
tunjid, line 47: It looks like a linter or formatter changed these when it didn't need to.

feature/compose/src/commonMain/kotlin/com/tunjid/heron/compose/ui/ComposeThreadGate.kt
```diff
@@ -24,36 +24,28 @@ import androidx.compose.ui.Modifier
...
-            threadGateSheetState.show(interactionSettingsPreference)
-        },
+        onClick = { threadGateSheetState.show(interactionSettingsPreference) },
     ) {
         PostInteractionStatus(
-            modifier = Modifier
```
tunjid, line 52: Same thing here.

tunjid:
In the feature modules you need to:

* Delete the unused action.
* Delete the methods for processing the unused action.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/threadgate/ThreadGateStateHolder.kt
```diff
@@ -0,0 +1,118 @@
...
+}
+
+sealed class ThreadGateAction(val key: String) {
+    data class Initialize(
+        val mode: Mode,
+        val allowed: ThreadGate.Allowed?,
+    ) : ThreadGateAction("Initialize")
+
+    data class UpdateAllowed(
+        val block: ThreadGate.Allowed.() -> ThreadGate.Allowed?,
```
tunjid: A data class with a lambda is a bit odd. Why not have the following?

```kotlin

    data class UpdateAllowed(
        val allowed: ThreadGate.Allowed?,
    )
```

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/threadgate/ThreadGateStateHolder.kt
```diff
@@ -95,12 +95,16 @@ private fun Flow<ThreadGateAction.Reset>.launchResetMutations(
 @Stable
 @Snapshottable
 interface ThreadGateState {
+    val recentLists: List<FeedList>
```
tunjid: The snapshottable plugin does this automatically, you don't need to redeclare interface fields. Declaring it once on the snapshot spec is enough. Gemini does not understand this.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/threadgate/ThreadGateStateHolder.kt
```diff
@@ -95,12 +95,16 @@ private fun Flow<ThreadGateAction.Reset>.launchResetMutations(
...
+    val mode: Mode?
+    val allowed: ThreadGate.Allowed?
+
     @SnapshotSpec
     @Serializable
     data class Immutable(
-        @Transient val recentLists: List<FeedList> = emptyList(),
-        @Transient val mode: Mode? = null,
-        @Transient val allowed: ThreadGate.Allowed? = null,
+        @Transient override val recentLists: List<FeedList> = emptyList(),
```
tunjid: We don't serialize the state here, so we can drop `@Serializable` and `@Transient`.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/threadgate/ThreadGateSheetState.kt
```diff
@@ -205,88 +181,95 @@ sealed class ThreadGateSheetState private constructor(
...
                     },
                 )
             }
 
-            Column(
-                verticalArrangement = Arrangement.spacedBy(0.dp),
-            ) {
-                val isCustomOrAnyone = !state.allowed.allowsNone
-
+            Column(verticalArrangement = Arrangement.spacedBy(0.dp)) {
```
tunjid, line 233: There's a few of these where your agent seems to remove deliberate line breaks and cause noisy diffs. Its fine for now, but I've added an agents.md file so hopefully it happens less.

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/threadgate/ThreadGateSheetState.kt
```diff
@@ -310,91 +291,49 @@ private fun ThreadGateBottomSheet(
                         text = stringResource(Res.string.thread_gate_select_from_your_lists),
                         style = MaterialTheme.typography.bodyMedium,
                     )
-
                     Icon(
-                        imageVector = if (listsExpanded)
-                            Icons.Rounded.KeyboardArrowUp
-                        else
-                            Icons.Rounded.KeyboardArrowDown,
-                        contentDescription = stringResource(
```
tunjid, line 319: Why was the content description removed?

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/threadgate/ThreadGateSheetState.kt
```diff
@@ -591,7 +530,7 @@ sealed class Mode {
     }
 }
 
-private val NoneAllowed = ThreadGate.Allowed(
+val NoneAllowed = ThreadGate.Allowed(
```
tunjid: Does this need to be public? Would internal have worked?

PR #1393, opened 2026-06-15
Migrate SelectedListSheetState to viewmodel backed state

joelmuraguri:
### What
- Migrates `SelectListSheetState` to the ViewModel-backed pattern

### Why
`recentLists` was previously threaded through `Filter` to `FilterLeaf` to `SocialListMemberFilter`, with `onUpdateRecentLists` dispatched from the
feature ViewModel to trigger loading. This coupled a generic, reusable picker sheet to feature-specific state and action plumbing.

### How

**ViewModel**
`SelectListViewModel` is `@AssistedInject`, `@Snapshottable` state, loads `recentLists` via `launchAndCollect` in the producer block same source (`RecordRepository.recentLists`) as `ThreadGateViewModel`.

**SheetState**
`SelectListSheetState` receives the ViewModel directly via the new`rememberBottomSheetState` overload. `lists` moves from a`mutableStateOf` field into ViewModel state, read via `produceState()`in the bottom sheet composable.

**`onListSelected` stays a callback**
`SelectListSheetState` is a shared picker used across multiple
features; what happens with the selected list differs per call site
(filter update, thread gate audience, etc.), so the ViewModel only owns loading; selection handling remains the feature's responsibility.

**Removed from the filter chain**
- `recentLists` and `onUpdateRecentLists` removed from `Filter`, `GrazeEditorViewModel` and `State`
`FilterLeaf`, and `SocialListMemberFilter`. 
- `SocialListMemberFilter`is now a `PaneScaffoldState` extension and creates its sheet state via `rememberSelectListSheetState`. `PaneScaffoldState` is threaded through
- `Filter` to `FilterLeaf` to support this.

**PaneScaffoldState extension**
```kotlin
fun PaneScaffoldState.rememberSelectListSheetState(
    onListSelected: (FeedList) -> Unit,
): SelectListSheetState
```

### Note
The list name display previously resolved via
`recentLists.find { it.uri.uri == filter.url }?.name` now falls back to
`filter.url` directly, since `recentLists` is no longer passed in.
Resolving the display name from the sheet's ViewModel state would be a follow-up if needed.

### DIFF
--- a/feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/GrazeEditorScreen.kt
+++ b/feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/GrazeEditorScreen.kt
@@ -41,7 +41,6 @@ import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
--- a/feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/GrazeEditorScreen.kt
+++ b/feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/GrazeEditorScreen.kt
@@ -48,7 +48,6 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.window.Dialog
 import androidx.compose.ui.window.DialogProperties
-import com.tunjid.heron.data.core.models.FeedList
 import com.tunjid.heron.data.core.models.Profile
 import com.tunjid.heron.data.graze.Filter
 import com.tunjid.heron.graze.editor.ui.filter.AnalysisFilter
@@ -151,10 +150,6 @@ fun GrazeEditorScreen(
                             animatedVisibilityScope = this@NestedNavigation,
                             paneScaffoldState = paneScaffoldState,
                             profileSearchResults = state.suggestedProfiles,
-                            recentLists = state.recentLists,
-                            onUpdateRecentLists = {
-                                actions(Action.UpdateRecentLists)
-                            },
                             filter = child,
                             atTopLevel = true,
                             onProfileQueryChanged = { query ->
@@ -235,7 +230,6 @@ fun GrazeEditorScreen(
 
     // Using Disposable effect as there's no need for a CoroutineScope
     DisposableEffect(hasList) {
-        if (hasList) actions(Action.UpdateRecentLists)
         onDispose { }
     }
 }
@@ -247,8 +241,6 @@ private fun Filter(
     paneScaffoldState: PaneScaffoldState,
     filter: Filter,
     profileSearchResults: List<Profile>,
-    recentLists: List<FeedList>,
-    onUpdateRecentLists: () -> Unit,
     atTopLevel: Boolean,
     path: List<Int>,
     enterFilter: (Int) -> Unit,
@@ -265,8 +257,6 @@ private fun Filter(
         modifier = modifier,
         filter = filter,
         profileSearchResults = profileSearchResults,
-        recentLists = recentLists,
-        onUpdateRecentLists = onUpdateRecentLists,
         index = index,
         path = path,
         onProfileQueryChanged = onProfileQueryChanged,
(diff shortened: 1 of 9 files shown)


feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/ui/filter/SocialFilters.kt
```diff
@@ -246,7 +241,7 @@ fun SocialListMemberFilter(
                 verticalArrangement = Arrangement.spacedBy(8.dp),
             ) {
                 OutlinedTextField(
-                    value = recentLists.find { it.uri.uri == filter.url }?.name ?: filter.url,
+                    value = filter.url,
```
tunjid: This is a regression. The previous code would show the name of the list if it was available, now it just shows a url.

feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/GrazeEditorScreen.kt
```diff
@@ -235,7 +230,6 @@ fun GrazeEditorScreen(
 
     // Using Disposable effect as there's no need for a CoroutineScope
     DisposableEffect(hasList) {
-        if (hasList) actions(Action.UpdateRecentLists)
         onDispose { }
     }
```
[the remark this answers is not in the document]
tunjid: This is valid

feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/GrazeEditorScreen.kt
```diff
@@ -539,14 +523,14 @@ fun FilterLeaf(
...
         )
-        is Filter.Social.ListMember -> SocialListMemberFilter(
-            modifier = modifier,
-            filter = filter,
-            recentLists = recentLists,
-            onUpdateRecentLists = onUpdateRecentLists,
-            onUpdate = onUpdate,
-            onRemove = onRemove,
-        )
+        is Filter.Social.ListMember -> with(paneScaffoldState) {
```
tunjid: Why use `with(paneScaffoldState) { ... }` as opposed to `paneScaffoldState.SocialListMemberFilter(...)`?

ui/timeline/src/commonMain/kotlin/com/tunjid/heron/timeline/ui/sheets/selectlist/SelectListStateHolder.kt
```diff
@@ -0,0 +1,62 @@
...
+            )
+        },
+    )
+
+context(productionScope: CoroutineScope)
+private fun launchLoadListsMutations(
+    state: SelectListState.SnapshotMutable,
+    recordRepository: RecordRepository,
+) = recordRepository.recentLists
+    .launchAndCollect { state.lists = it }
```
tunjid: You will need to rebase on the latest main, `` has been deleted from the codebase. The mutator library's version is called `launchedCollect { ... }`

feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/GrazeEditorScreen.kt
joelmuraguri: I've made a fix on this

feature/graze-editor/src/commonMain/kotlin/com/tunjid/heron/graze/editor/ui/filter/SocialFilters.kt
```diff
@@ -208,22 +209,22 @@ fun SocialStarterPackFilter(
...
     val sheetState = rememberSelectListSheetState(
-        lists = recentLists,
         onListSelected = { list ->
-            onUpdate(
-                filter.copy(url = list.uri.uri),
-            )
+            onUpdate(filter.copy(url = list.uri.uri))
         },
     )
+    val selectListState = sheetState.viewModel.produceState()
```
tunjid: Lets not use `produceState()` here, we should also make the `viewModel` on the `sheetState` internal.

* `produceState()` will cause state production even when the sheet is not visible, which we don't want. We can just read the state with sheetState.state. You will need to add a getter to read the state from the viewModel in the sheetState.
* The viewmodel should be an internal impl detail. At some point we may want to use the `retainedState { }` APIs instead of a viewmodel.
* `produceState` is already called inside the `ModalBottomSheet { }` this is what cause state production to occur only when the sheet is visible.

tunjid (approved):

PR #1404, opened 2026-06-22
feat: add toggle to hide replies from people you don't follow

### LINKED ISSUE #1354: Add option to filter replies by follows
When Show Replies is enabled for your Following feed, it would be nice to have an option to filter out replies to people that you don't also follow. Deck.blue has this option and it works really well.

mvanhorn:
## Summary
Adds a "Hide replies from people you don't follow" toggle to the Following feed preferences. When enabled alongside "Show Replies", replies addressed to accounts the signed-in user does not follow are filtered out of the feed, matching the behavior requested in #1354.

## Why this matters
The `FeedPreference` data model already carried a `hideRepliesByUnfollowed` field that was synced through the preference updater but never surfaced in the UI nor applied to the feed query. This change exposes that existing field:

- Adds a `shouldHideRepliesByUnfollowed` companion accessor mirroring `shouldHideReplies`.
- Adds a new `SettingsToggleItem` in `FeedPreferencesSection` bound to `feedPreference.copy(hideRepliesByUnfollowed = it)` with a new `timeline_preferences_replies_by_unfollowed` string.
- Threads a `hideRepliesByUnfollowed` parameter from the `TimelineRepository.feedItems(...)` call sites into `TimelineDao.feedItems(...)`, and extends the DAO `@Query` so a reply is kept only when its parent post author is followed (an `EXISTS` subquery joining `posts.uri = timelineItems.parentPostUri` to `profileViewerStates.following`).
- Scopes the filter to the following feed: non-following sources explicitly pass `hideRepliesByUnfollowed = false`, since the data-class default for this field is `true`.

The setting flows end to end: toggle in `NestedFeedPreferences` -> `FeedPreference.hideRepliesByUnfollowed` -> `shouldHideRepliesByUnfollowed` accessor -> `TimelineRepository.feedItems` -> `TimelineDao.feedItems` `@Query` filter, and persists through the existing preference updater sync.

## Testing
Manual verification against the test scenarios in the issue:
- Toggle off (default for non-following sources): replies from unfollowed authors remain visible.
- Toggle on with "Show Replies" enabled: replies to followed users still show, replies to unfollowed users are filtered out.
- Top-level (non-reply) posts are unaffected regardless of toggle state.
- Preference round-trips through the preference updater sync and persists across restart.

No schema migration is required: the change only adds a query parameter and SQL clause over the existing `timelineItems`, `posts`, and `profileViewerStates` tables. CI will validate the full Kotlin Multiplatform build.

Fixes #1354

### DIFF
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/TimelineRepository.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/TimelineRepository.kt
@@ -42,6 +42,7 @@ import com.tunjid.heron.data.core.models.FeedPreference
 import com.tunjid.heron.data.core.models.FeedPreference.Companion.homeFeedOrDefault
 import com.tunjid.heron.data.core.models.FeedPreference.Companion.shouldHideQuotes
 import com.tunjid.heron.data.core.models.FeedPreference.Companion.shouldHideReplies
+import com.tunjid.heron.data.core.models.FeedPreference.Companion.shouldHideRepliesByUnfollowed
 import com.tunjid.heron.data.core.models.FeedPreference.Companion.shouldHideReposts
 import com.tunjid.heron.data.core.models.Post
 import com.tunjid.heron.data.core.models.Preferences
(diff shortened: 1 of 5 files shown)


tunjid:
I need to rebase this on the current main, and then address the PR feedback before merging. I'm closing this now to do that, thanks for the contribution.

tunjid:
Upstreamed in https://github.com/tunjid/heron/pull/1429

PR #1444, opened 2026-07-04
fix: launch playlist diffing coroutine lazily instead of in init

joelmuraguri:
## Summary
- Fixes the TODO on `ExoplayerController`'s diffing coroutine: it was launched eagerly in `init`, meaning it started running as soon as an
`ExoplayerController` was constructed, before any video was ever registered or played.

## Change
- `diffingJob` is now launched with `CoroutineStart.LAZY` in `init`,
  so the coroutine is built but not scheduled to run.
- `registerVideo`  the first call site that can ever send into `mediaItemMutationsChannel`  now calls `diffingJob?.start()` before
  doing anything else. `Job.start()` is idempotent, so this is safe to call on every `registerVideo` invocation.
- Removed a stray `diffingJob?.cancel()` that preceded the job's only assignment (dead code, `diffingJob` was always `null` there).

## Why this matters
- `mediaItemMutationsChannel` is a rendezvous channel (`Channel<...>()`, capacity 0). This makes `.start()` a hard requirement, not a nice-to-have:if the job were marked lazy without ever calling `.start()`, the first
- `send` in `registerVideo` would suspend forever with nobody collecting,
silently dropping the video from the playlist. This PR wires up both
halves of that contract.

- `teardown()`'s `diffingJob?.cancel()` needs no changes  cancelling a
job that was never started is already safe.

## Out of scope
- The second TODO in the same `init` block (the two `snapshotFlow{...}.launchIn(...)` flows for pause-on-idle and mute) is a related but separate cleanup. `launchIn` has no lazy variant, so making those lazy requires rewriting them as plain `scope.launch(start
= CoroutineStart.LAZY) { flow.collect {...} }`, which felt like a separate, slightly bigger change. Happy to follow up with that in a
second PR if desired.

## Testing
- Verified `registerVideo` `play` still populates and plays the ExoPlayer playlist as before.
- Verified calling `registerVideo` multiple times does not create multiple diffing collectors (`Job.start()` no-ops after the first call).
- Verified `teardown()` immediately after construction (no video ever registered) cancels cleanly with no exceptions.

### DIFF
--- a/ui/media/src/androidMain/kotlin/com/tunjid/heron/media/video/ExoplayerController.kt
+++ b/ui/media/src/androidMain/kotlin/com/tunjid/heron/media/video/ExoplayerController.kt
@@ -44,6 +44,7 @@ import dev.andrewbailey.diff.differenceOf
 import java.io.File
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.channels.Channel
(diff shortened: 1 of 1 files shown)


tunjid:
I was hoping to remove the `init` block completely with this. The idea is for some coroutine scope to actually own the diffing.

Issue #1471, opened 2026-07-09
Post with image does not post


Penworks:
**Describe the bug**
A post with short text plus image from camera roll fails. 

**To Reproduce**
1. Create a new post
2. Add photo from camera folder
3. Click publish

**Expected behavior**
Instead of a new post appearing, the post gets stuck in Fail. Several retry's make no difference. 

**Screenshots**

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
 - Device: Moto Edge Neo
 - OS: Android 16
 - Browser Brave/Vivaldi 
 - Version up to date

**Additional context**
- This error has happened several times. 
- It may be due to file size of high quality photos? I have made the same posts to bluesky using other apps (Tokimeki, native) and they publish immediately.

tunjid:
I'm so sorry about that. I think it's a network timeout on my side. I'll fix for the next release.

tunjid:
This should be fixed in the next release.

Penworks:
Hey tx for messages. No problem I had the error, it's a new app, really nice. I can join beta release if you have it. 

I'll update and test and come back if I still get errors on this phone.

PR #1473, opened 2026-07-09
Update HttpTimeout configurations for media uploads

tunjid:


tunjid:
Fixes https://github.com/tunjid/heron/issues/1471

Issue #1210, opened 2026-04-14
Following feed reaches an "end"


nick-mcmanus:
**Describe the bug**
The default Following feed reaches an "end" and stops loading more posts.

**To Reproduce**
Use the default Following feed, keep scrolling and you'll eventually reach an end where it stops loading more posts.

**Expected behavior**
It should keep loading more posts indefinitely.

**Smartphone (please complete the following information):**
 - Device: Pixel 9 Pro
 - OS: Android 16
 - Version: 2.2.3-bugfix-2.2.3

HI-JIN2:
I'd like to look into this issue.

It may be related to pagination.

HI-JIN2:
`Home > Discover `scrolls infinitely, but `Home > Following` does not.

HI-JIN2:
Following and Discover do not use the same API.

The Home > Following tab uses `app.bsky.feed.getTimeline`.
Saved/custom feeds such as Discover use `app.bsky.feed.getFeed`.

Because of that, it seems possible for Discover to keep receiving cursors normally while only Following stops paginating.

HI-JIN2:
@tunjid  I checked this against my Bluesky Following feed, and Heron shows the same result.

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/c466c17c-1d7b-4168-8514-082bb1cc9183" />

In both cases, the feed eventually ends with the message: "You've reached the end of your feed! Find some more accounts"

So this does not seem to be a Heron-only pagination issue. It looks like Following from the official API itself reach an end, rather than scrolling indefinitely.


Given that, how would you like to handle this issue?

A couple of possible options might be:
- Show an end-of-feed message in Heron, similar to Bluesky
- Implement a custom infinite scroll behavior on Heron’s side instead of following the official API behavior as-is

nick-mcmanus:
In my case, it would stop loading posts in Following around the 4 hour mark (posts themselves were 4 hours old). Whereas the Bluesky app would keep loading posts past that.

It may have been a one-off bug that I experienced only twice in Heron. Checking just now, I'm able to go back and see posts in Following from days ago without issue.

It should eventually reach an end I imagine, but not like I saw in the aforementioned bug experiences.

tunjid:
This is more likely due to pagination getting stuck bc of a network cursor that returned an empty list. We probably need a way to "jog" the pagination pipeline

nick-mcmanus:
[...]

PR #1529, opened 2026-08-02
Set cursor for initial quieries to better anchor timelines

tunjid:


tunjid:
Fixes https://github.com/tunjid/heron/issues/1210

PR #1607, opened 2026-08-31
Improve video upload reliability and QoL with a persistent notification

karanshukla:
Hi @tunjid !

Over the weekend I ran into an issue where video uploads failed from my Pixel 9 Pro. Digging into it, I realised it may have been due to the OS removing the task whenever I swiped away from the app or locked the phone. Based off my digging, I figured the best first pass way of solving it would be to introduce a persistent notification that holds the task alive. Upstream in the AT Protocol they also enabled 8+ minute videos, so I'm certain this is something that would occur for longer videos.

The video which I eventually managed to upload is here: [Video](https://bsky.app/profile/karanshukla.ca/post/3mudiuha5yk44)

~~As a QoL improvement I also fixed media uploads when signing in via an app password, as that sign in method doesn't return the PDS URL, only OAuth does. This is necessary for testing locally with a dev build as the dev builds don't support OAuth login.~~

### Design Decisions

I made a couple of design decisions here that I wanted some input into, so feel free to update or change anything as you prefer!

1. New Notification channel for uploads (heron.uploads). This is forked off heron.transfers
2. stat_sys_uploads icon for uploads. 
3. "Uploading Media" string for the notification

### iOS implementation

I don't have a personal Apple device so I can't implement an iOS version without verification, but we can use the Background Task Scheduler, which I found out about here: https://medium.com/@kmpbits/sleeping-but-working-cross-platform-background-sync-with-kmp-70811e1dbd90

This required me to abstract out a task and a hold so that Android and iOS had separate implementations, but the effect is the same. `BackgroundTaskScheduler` in `commonMain` keeps the app-facing API the same (`enqueue`, `status`, `cancel`) but iOS gets its own background worker for video uploads without the notification.


data/core/src/commonMain/kotlin/com/tunjid/heron/data/network/VideoUploadService.kt
```diff
@@ -99,7 +100,9 @@ internal class SuspendingVideoUploadService(
             .filterNotNull()
             .first()
 
-        val serviceUrl = authToken.serviceUrl ?: throw IllegalStateException("Not signed in")
+        val serviceUrl = authToken.serviceUrl
+            ?: identityResolver.resolvePds(Did(signedInProfileId.id))?.toString()
+            ?: throw IllegalStateException("Could not resolve a PDS for ${signedInProfileId.id}")
```
karanshukla: This exception is unlikely to actually throw but best to handle it anyway.

data/tasks/src/androidMain/kotlin/com/tunjid/heron/data/tasks/TransferRunner.kt
```diff
@@ -58,3 +78,40 @@ internal suspend fun Context.runTransfer(
...
+    ?: TransferNotice(
+        channelId = Task.Kind.Transfer.channelId,
+        title = DownloadTitle,
+        smallIcon = android.R.drawable.stat_sys_download,
+        progress = null,
+    )
+
+private const val DownloadTitle = "Download"
+
+private const val UploadTitle = "Uploading media"
```
karanshukla: Unsure if this is the correct way to handle strings in the project but please let me know!

tunjid:
Hi! 

* The app password service url fix is right, ty for that.
* For using the background service API for media uploads, the current API isn't robust enough for it yet largely bc iOS and Android share the same API shape for background work in the app, but have different implementations. While your fix works on Android, it'll break iOS, though desktop might be fine. Its actually the reason uploads are currently done in the foreground.

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/writequeue/WriteQueue.kt
```diff
@@ -250,6 +258,29 @@ internal class PersistedWriteQueue(
...
 
+    private suspend fun Writable.keepingUploadAlive(
+        write: suspend () -> Outcome,
+    ): Outcome {
+        if (!carriesMedia()) return write()
+        val taskId = TaskId("$UploadTaskPrefix$queueId")
+        // The upload still runs without the keep alive; the OS is just free to kill it early.
+        runCatchingUnlessCancelled {
+            backgroundTaskScheduler.cancel(taskId)
+            backgroundTaskScheduler.enqueue(Task.Upload(id = taskId))
```
tunjid: For this, it seems the task scheduler doesn't actually run the upload. It instead monitors the upload queue for the scheduled work to be drained.

Its a really clever solution for Android, however it requires that the app have been opened first. Draining the queue is done in the UI: https://github.com/tunjid/heron/blob/main/ui/scaffold/src/commonMain/kotlin/com/tunjid/heron/ui/scaffold/scaffold/AppScaffoldState.kt#L278

So if Android workManager reschedules the work after the app has been killed, it will hang indefinitely bc there's no UI to actually perform the drain.

The real fix is to drain the queue independent from the UI but I haven't thought up a good abstraction for this across Android, iOS and Desktop. Feel free to give that a go, however this implementation may cause the app to run out of it's the background work quota.

karanshukla:
@tunjid I'll split off the PDS fix into a separate PR shortly and take a look at your other feedback, thanks!

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/writequeue/WriteQueue.kt
karanshukla: Thanks for the heads up, I didn't think of the iOS compatibility since I have no device to test with. But I agree, having an Android/desktop only solution isn't ideal.

I'll take a look at draining the queue independently but I'm new to native app development so I want to be careful with it. I'll move this PR back to draft and open up the PDS issue separately.

karanshukla:
@tunjid tried out a couple of options, let me know what you think, it involved the abstraction you described, with separating out the task and holds so that we could get iOS its own implementation. I don't have a way to test with iOS but I followed the guide here: https://developer.apple.com/documentation/backgroundtasks

tunjid:
Hi! I'll take a look again tomorrow :)

data/core/src/commonMain/kotlin/com/tunjid/heron/data/utilities/writequeue/WriteQueue.kt
```diff
@@ -240,8 +243,10 @@ internal class PersistedWriteQueue(
         emit(
             Pair(
                 first = writable,
-                second = with(writable) {
-                    withTimeout(writable.writeTimeout()) { write() }
+                second = writable.keepingUploadAlive {
```
tunjid, line 246: `keepingUploadAlive` still seems to launch the background task as a sibling of the write, rather than using the underlying OS to actually perform the write.

It is better for writes with media to start a task that performs the write in the task's process. Right now it launches a task that may be in a separate process, and keeps it around till the write completes. It only works bc right now the task coincidentally shares the same process as the regular app.

karanshukla: Yep, i see the issue here with that. With the APIs, its difficult to introduce a background task that isnt just a floating process disconnected from the main app.  

In doing some research I discovered we probably want to use the UIDT API (https://developer.android.com/develop/background-work/background-tasks/uidt) but I dont see an easy way of implementing this with the current code arch. I think its best to close this PR out unless you're willing to go down the route of rewriting some of the data arch to support UIDT cleanly without breaking iOS

PR #1620, opened 2026-09-04
Pick like/repost notification differentation from the push reason

### LINKED ISSUE #1263: Push notification shows repost even if its a regular post
**Describe the bug**
When receiving push notifications of interactions of posts, it shows reposts rather than posts.

**To Reproduce**
1. Get a push notification

**Expected behavior**
It shows posts rather than reposts

**Screenshots
Image
Image

**Smartphone (please complete the following information):**
 - Device: Xiaomi Redmi Note 13 Pro 5G
 - OS: LineageOS Android 16

karanshukla:
Fixes #1263. Push notifications told post authors "liked your repost" about posts they wrote and never reposted.

This occurred due to the way the app view works. When the App View emits the records for the likes, Heron resolves the like event created by the person who liked the repost. As a result, the person who made the original post receives a notification indicating their repost was liked. 

Fix is to use the notification Reason instead of trusting the resolved record blindly.

## Misc notes

- `resolvePushNotification` has no test coverage.
- The push path hardcodes `reasonSubject = null` at every construction site, so could be an opportunity for a future refactor

### DIFF
--- a/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
+++ b/data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
@@ -65,7 +65,6 @@ import com.tunjid.heron.data.core.types.NotificationFilteredOutException
 import com.tunjid.heron.data.core.types.PostUri
 import com.tunjid.heron.data.core.types.ProfileId
 import com.tunjid.heron.data.core.types.RecordUri
-import com.tunjid.heron.data.core.types.RepostUri
 import com.tunjid.heron.data.core.types.RestrictedProfileException
 import com.tunjid.heron.data.core.types.UnknownNotificationException
 import com.tunjid.heron.data.core.types.profileId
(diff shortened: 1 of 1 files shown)


data/core/src/commonMain/kotlin/com/tunjid/heron/data/repository/NotificationsRepository.kt
```diff
@@ -555,8 +554,8 @@ internal class OfflineNotificationsRepository(
                         // Like or Like via repost
                         is Like -> when (resolvedRecord.post.viewerStats?.threadMuted) {
                             true -> throw MutedThreadException(resolvedRecord.post.uri)
-                            else -> when (resolvedRecord.via) {
```
karanshukla, line 558: This is where things got awry, the resolved record would always be the repost like.

tunjid: Great catch, ty

tunjid (approved):

